PR #32, opened 2023-10-12
proxy: Implement prototype

### LINKED ISSUE #13: proxy: Create first draft

jakoschiko:
First prototype for an IMAP proxy.

It can be run with:

```sh
cd proxy
RUST_LOG=proxy=trace cargo run
```

There is a default config file located at `proxy/config.toml`. It contains ports that can be used to connect with a local email client.

The change set contains some TODOs. We need to create issues for them after merging this PR.

Closes #13

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,10 @@ tokio = { version = "1.32.0", features = ["io-util"] }
 
 [dev-dependencies]
 tokio = { version = "1.32.0", features = ["macros", "net", "rt"] }
+
+[workspace]
+resolver = "2"
+members = [
+  ".",
+  "proxy"
+]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,5 @@ tokio = { version = "1.32.0", features = ["macros", "net", "rt"] }
 [workspace]
 resolver = "2"
 members = [
-  ".",
   "proxy"
 ]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -13,3 +13,10 @@ tokio = { version = "1.32.0", features = ["io-util"] }
 
 [dev-dependencies]
 tokio = { version = "1.32.0", features = ["macros", "net", "rt"] }
+
+[workspace]
+resolver = "2"
+members = [
+  ".",
```
duesee, line 20:
```suggestion
```

If I understand correctly, line 20 is not needed. See https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package

duesee (approved):
Awesome, thanks!

duesee:
The build fails because `toml_edit` quires Rust 1.66 but we enforce 1.65 in the CI. See https://github.com/duesee/imap-flow/blob/main/.github/workflows/build_and_test.yml#L103.

I don't have a strong opinion how to resolve this. We could just increase the MSRV? If you would like to do it here, you could also change `imap-codec` -> `imap-flow`.

jakoschiko:
> If you would like to do it here, you could also change `imap-codec` -> `imap-flow`.

Done

duesee:
Thanks! I added a commit to also install `libssl-dev` in the `minimal-versions` job. Not sure if it does the trick.

duesee:
Actually, we don't need `native-tls` (and `openssl`) at all. Reverted and removed the dependency.

PR #37, opened 2023-10-16
Refactor `imap_flow::stream`

### LINKED ISSUE #18: Find better way for consuming AsyncRead + AsyncWrite + Send types
- The constructors of ClientFlow and ServerFlow take byte stream. This stream needs to implement AsyncRead + AsyncWrite + Send
- We achieve by using several types, see module imap-flow::stream
- Maybe there is a more lightweight way for achieving this with a smaller API

jakoschiko:
Previously there were two public types in `imap_flow::stream`:

- The trait `Stream`: This trait existed only for technical reasons. The user might think that they should implement it, which is wrong.
- The struct `AnyStream`. Handles the boxing, the pinning and the dynamic dispatch. This is the type that the user should interact with.

The new API looks like this:
- `imap_flow::stream` has now only a single type, the struct `Stream`
  - It can be constructed with `_::new`
  - It provides access to the underlying `AsyncRead` via `_::read` and `_::read_mut`
  - It provides access to the underlying `AsyncWrite` via `_:::write` and `_::write_mut`
- `ClientFlow` and `ServerFlow` provide access to the underlying `Stream` via `_::stream` and `_::stream_mut`

Closes #18

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -2,16 +2,52 @@ use std::pin::Pin;
 
 use tokio::io::{AsyncRead, AsyncWrite};
 
-// TODO: Reconsider this. Do we really need Stream + AnyStream? What is the smallest API that we need to expose?
+trait AsyncReadWrite: AsyncRead + AsyncWrite + Send {
+    fn read(self: Pin<&Self>) -> Pin<&(dyn AsyncRead + Send)>;
+    fn read_mut(self: Pin<&mut Self>) -> Pin<&mut (dyn AsyncRead + Send)>;
+    fn write(self: Pin<&Self>) -> Pin<&(dyn AsyncWrite + Send)>;
+    fn write_mut(self: Pin<&mut Self>) -> Pin<&mut (dyn AsyncWrite + Send)>;
+}
 
-pub trait Stream: AsyncRead + AsyncWrite + Send {}
+impl<S: AsyncRead + AsyncWrite + Send> AsyncReadWrite for S {
+    fn read(self: Pin<&Self>) -> Pin<&(dyn AsyncRead + Send)> {
+        self
+    }
+    fn read_mut(self: Pin<&mut Self>) -> Pin<&mut (dyn AsyncRead + Send)> {
+        self
+    }
+    fn write(self: Pin<&Self>) -> Pin<&(dyn AsyncWrite + Send)> {
+        self
+    }
+    fn write_mut(self: Pin<&mut Self>) -> Pin<&mut (dyn AsyncWrite + Send)> {
+        self
+    }
+}
 
-impl<S: AsyncRead + AsyncWrite + Send> Stream for S {}
+pub struct Stream {
+    underlying: Pin<Box<dyn AsyncReadWrite>>,
+}
 
-pub struct AnyStream(pub Pin<Box<dyn Stream>>);
+impl Stream {
+    pub fn new<S>(stream: S) -> Self
+    where
+        S: AsyncRead + AsyncWrite + Send + 'static,
+    {
+        Self {
+            underlying: Box::pin(stream),
+        }
+    }
 
-impl AnyStream {
-    pub fn new<S: Stream + 'static>(stream: S) -> Self {
-        Self(Box::pin(stream))
+    pub fn read(&self) -> Pin<&(dyn AsyncRead + Send)> {
+        self.underlying.as_ref().read()
+    }
+    pub fn read_mut(&mut self) -> Pin<&mut (dyn AsyncRead + Send)> {
+        self.underlying.as_mut().read_mut()
+    }
+    pub fn write(&self) -> Pin<&(dyn AsyncWrite + Send)> {
+        self.underlying.as_ref().write()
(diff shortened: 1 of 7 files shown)


jakoschiko:
I'm not sure if this is really the best solution. We should discuss this.

src/stream.rs
```diff
@@ -2,16 +2,52 @@ use std::pin::Pin;
 
 use tokio::io::{AsyncRead, AsyncWrite};
 
-// TODO: Reconsider this. Do we really need Stream + AnyStream? What is the smallest API that we need to expose?
+trait AsyncReadWrite: AsyncRead + AsyncWrite + Send {
+    fn read(self: Pin<&Self>) -> Pin<&(dyn AsyncRead + Send)>;
+    fn read_mut(self: Pin<&mut Self>) -> Pin<&mut (dyn AsyncRead + Send)>;
+    fn write(self: Pin<&Self>) -> Pin<&(dyn AsyncWrite + Send)>;
+    fn write_mut(self: Pin<&mut Self>) -> Pin<&mut (dyn AsyncWrite + Send)>;
```
duesee, line 9: Question: Should we name these `reader`, `reader_mut`, `writer`, `writer_mut`, instead?

duesee (approved):
Thanks!

Recapping to test my understanding:

The main improvement is that instead of `pub trait Stream`, we now have `trait AsyncReadWrite` (which doesn't need to be exposed.) This is possible by introducing the `read`, `read_mut`, ... methods that don't expose a trait pairing `AsyncRead + AsyncWrite` but gives separate handles `AsyncRead`, `AsyncWrite`.

Technically, we could also "inline" the `AsyncReadWrite` trait and `impl` all methods directly on `Stream`?

Are there any downsides to this solution despite it having more code?

Bonus:
* Better names `AnyStream -> Stream` ✅ 
* Better access `.0 -> write_mut` (see question) ✅ 

From my understanding, this PR is definitely an improvement. As we are still in an experimenting phase, I would propose to merge it after resolving the rename suggesstion. Then, I will rebase my work on the examples + tui and we can see how it goes.

jakoschiko:
> Technically, we could also "inline" the `AsyncReadWrite` trait and `impl` all methods directly on `Stream`?

This is not possible because dyn can only be combined with a single (normal) trait (yet).

Similarly dyn downcasting is not (yet) possible, so we need these methods on the trait.

Another option would be to implement `AsyncRead` and `AsyncWrite` for `Stream`. In that cause we could get rid of these `read`, ... methods. I already tried it and it was annoying because I needed to write a lot of delegating code. But it made the API even smaller, so we could discuss it.

jakoschiko:
> We could also get rid of `Stream` and dynamic dispatch altogether. Start with `ClientFlow<S>` and propagate the `AsyncRead + AsyncWrite + Send + Unpin` requirements. I just followed the compiler errors w/o thinking too much :-)

This type parameter is parasitic and will effect many types. It will result in very, VERY ugly type annotations... e.g.

```rust
Proxy<ConnectedState<TcpStream, TlsStream<TcpStream>>>
```

But hey, that's how Rust works, it's worth a try.

> PS: Urgs, STARTTLS may require dyn :-/ Or IMAP's `COMPRESS` extension.

Can you elaborate this more?

jakoschiko:
> ... and propagate the `AsyncRead + AsyncWrite + Send + Unpin` requirements ...

Are you sure about `Unpin`? I think it's not necessary if we do `Box::pin(stream)` inside the constructor of `ClientFlow` and `ServerFlow`.

duesee:
> Are you sure about Unpin?

No. It was just a draft to see how bad it would be :-)

> Can you elaborate this more?

When STARTTLS is used, we need to replace `TcpStream` with `TlsStream` w/o closing the connection. This is easy with `Box<dyn Stream>` but requires extra work w/o dynamic dispatch.

When COMPRESS is used, we need to replace `TcpStream` with `DeflateStream` w/o closing the connection. 

Combined... we need to handle (and make room for) ...

* `TcpStream`,
* `DeflateStream(TcpStream)`,
* `TlsStream(TcpStream)`,
* `TlsStream(DeflateStream(TcpStream))`

... in our client.

`Box<dyn Stream>` sounds most ergonomic.

jakoschiko:
Interesting... I'm not sure how we should proceed. Maybe we should keep dyn trait. Maybe we should avoid both dyn trait and the  type parameter and instead be more strict so we can prevent cases like `DeflateStream<DeflateStream<TcpStream>>`. I suggest to postpone this issue and continue with the other features first. Future knowledge might improve our decisions on this topic.

duesee:
Okay! I feel that an enum with four states could be a bit annoying. But it would definitely be the tightest model of a sane IMAP reality.

Feel free to close this PR. Looking forward to revisit!

duesee:
This gets out-of-sync. Shall we close for now? Maybe there are fragments we could still merge?

jakoschiko:
I would like to keep this PR for irrational reasons.

duesee:
Release it. It wants to be free!

PR #43, opened 2023-10-18
docs: Provide simple client and server examples

duesee:
I initially wanted to copy the examples from `imap` and `async_imap`. However, in order to do so, we already need a higher-level API. It doesn't make a lot of sense to implement these examples now. So, for now, I'll use the examples as playground for just the things we want to abstract over in imap-flow, i.e., literal handling, authentication, and idle, and be careful to not do too much already.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -1,0 +1,43 @@
+use imap_codec::imap_types::{
+    command::{Command, CommandBody},
+    core::Tag,
+};
+use imap_flow::{
+    client::{ClientFlow, ClientFlowEvent, ClientFlowOptions},
+    stream::AnyStream,
+};
+use tokio::net::TcpStream;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let stream = TcpStream::connect("127.0.0.1:12345").await.unwrap();
+
+    let (mut client, greeting) =
+        ClientFlow::receive_greeting(AnyStream::new(stream), ClientFlowOptions::default())
+            .await
+            .unwrap();
+    println!("received greeting: {greeting:?}");
+
+    let handle = client.enqueue_command(Command {
+        tag: Tag::try_from("A1").unwrap(),
+        body: CommandBody::Noop,
+    });
+
+    loop {
+        match client.progress().await.unwrap() {
+            ClientFlowEvent::CommandSent {
+                tag,
+                handle: got_handle,
+            } => {
+                println!("command sent: {tag:?}, {handle:?}");
+                assert_eq!(handle, got_handle);
+            }
+            ClientFlowEvent::DataReceived { data } => {
+                println!("data received: {data:?}");
+            }
+            ClientFlowEvent::StatusReceived { status } => {
+                println!("status received: {status:?}");
+            }
+        }
+    }
+}
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -1,0 +1,42 @@
+use imap_codec::imap_types::response::{Greeting, Status};
+use imap_flow::{
+    server::{ServerFlow, ServerFlowEvent, ServerFlowOptions},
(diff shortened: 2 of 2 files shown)


examples/client.rs
```diff
@@ -0,0 +1,39 @@
...
+    println!("received greeting: {greeting:?}");
+
+    client.enqueue_command(Command {
+        tag: Tag::try_from("A1").unwrap(),
+        body: CommandBody::Noop,
+    });
+
+    loop {
+        match client.progress().await.unwrap() {
+            ClientFlowEvent::CommandSent { tag, handle } => {
```
duesee, line 28: It's a bit weird to have `tag` and `handle` here. Ideally, the handle can be used to obtain the command. And when this can be done, the `tag` can be obtained through the obtained command.

I stumbled upon this because I wanted to `Debug` print the sent command. So, I needed to use an external `HashMap<Handle, Command>`. But after `let cmd = map.remove(handle)`, I can already access the tag through `cmd`.

Maybe there should be a `handle.into_command()` or so?

examples/server.rs
```diff
@@ -0,0 +1,37 @@
...
+    .unwrap();
+
+    loop {
+        match server.progress().await.unwrap() {
+            ServerFlowEvent::CommandReceived { command } => {
+                println!("command received: {command:?}");
+                server.enqueue_status(Status::no(Some(command.tag), None, "...").unwrap());
+            }
+            ServerFlowEvent::ResponseSent { handle } => {
+                println!("response sent: {handle:?}");
```
jakoschiko, line 33: I assume `ResponseSent` might also contain the response in the future.

examples/client.rs
```diff
@@ -0,0 +1,39 @@
...
+async fn main() {
+    let stream = TcpStream::connect("127.0.0.1:12345").await.unwrap();
+
+    let (mut client, greeting) =
+        ClientFlow::receive_greeting(AnyStream::new(stream), ClientFlowOptions::default())
+            .await
+            .unwrap();
+    println!("received greeting: {greeting:?}");
+
+    client.enqueue_command(Command {
```
jakoschiko, line 21: `enqueue_command` returns a handle. If you assign it to a val and print it, then it might be more clear to the reader what the purpose of the handle inside `CommandSent` is.

This could be important because the handle is not IMAP concept, so even a developer that knows IMAP might not immediately understand its meaning.

examples/server.rs
```diff
@@ -0,0 +1,37 @@
...
+        Greeting::ok(None, "Hello, World!").unwrap(),
+    )
+    .await
+    .unwrap();
+
+    loop {
+        match server.progress().await.unwrap() {
+            ServerFlowEvent::CommandReceived { command } => {
+                println!("command received: {command:?}");
+                server.enqueue_status(Status::no(Some(command.tag), None, "...").unwrap());
```
jakoschiko, line 30: Same for the handle returned by `enqueue_status`.

jakoschiko (approved):
I really like how simple the examples look like.

examples/client.rs
duesee: Yeah. It's also not entirely clear to me how to make good use of the handle.

The loop could, in theorie, yield one or more received events before the sent event. Are we expected to take this into account? E.g., to diverge our logic into pre- and post arrival of the sent event?

Are we expected to verify that the thing we `enqued` was really sent? Or do we implicitly assume that our command was processed because we got a corresponding command completion result (status with same tag as command)?

duesee: These were the questions I had when implementing a draft version of `client.login()`, `client.fetch()`, etc.

Spoiler: It all works amazingly well so far. But I mostly just ignore `SentCommand`.

duesee: Resolving for now. Will open an issue.

PR #51, opened 2023-10-24
feat(proxy): Implement minimal argument processing

duesee:
Requesting a review due to the additional dependency. Maybe you have another preference. But `argh`seems to be a very small dependency (and I like the concept).

### DIFF
--- a/proxy/.gitignore
+++ b/proxy/.gitignore
@@ -1,2 +1,4 @@
 # direnv (https://direnv.net/)
 .envrc
+
+private
--- a/proxy/Cargo.toml
+++ b/proxy/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0.75"
+argh = "0.1.12"
 colored = "2.0.4"
 imap-codec = { version = "1.0.0", features = ["bounded-static", "quirk_crlf_relaxed"] }
 imap-flow = { path = ".." }
--- a/proxy/src/main.rs
+++ b/proxy/src/main.rs
@@ -3,11 +3,20 @@ mod proxy;
 mod util;
 
 use anyhow::{Context, Result};
+use argh::FromArgs;
 use config::{Config, Service};
 use proxy::{ClientAcceptedState, Proxy};
 use tokio::task::JoinSet;
 use tracing::{error, instrument, Instrument};
 use tracing_subscriber::EnvFilter;
+
+/// IMAP proxy.
+#[derive(FromArgs)]
+struct Arguments {
+    /// optional config path ("config.toml" by default)
+    #[argh(option, default = "String::from(\"config.toml\")")]
+    config: String,
+}
 
 #[tokio::main]
 async fn main() -> Result<()> {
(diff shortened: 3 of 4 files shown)


proxy/.gitignore
```diff
@@ -1,2 +1,4 @@
 # direnv (https://direnv.net/)
 .envrc
+
+private
```
duesee, line 4: On my machine, I put an copy of the old `config.toml` here and will use `--config private/config.toml` from now on. You could do the same. But I think we do not want actual providers in our default `config.toml`.

jakoschiko: > On my machine, I put a copy of the old `config.toml` (with the providers I use) here and will use `--config private/config.toml` from now on. 

Doesn't git allow a global `.gitignore` in the home directory for such use cases? Otherwise you would need to add personal stuff that is not related to the project (IDE config, etc.) to all local `.gitignore` files.

> You could do the same. But I think we do not want actual providers in our default `config.toml`.

Why not? It's convenient. IMO convenience is more important than cleanness here.

proxy/src/main.rs
```diff
@@ -20,13 +29,10 @@ async fn main() -> Result<()> {
         .without_time()
         .init();
 
-    // Load config file
-    let config = {
-        let config_path = std::env::args().nth(1).unwrap_or("config.toml".to_string());
+    let args: Arguments = argh::from_env();
```
jakoschiko, line 32: I'd keep comments like `// Load command arguments` and `// Load proxy config` because then the main function is readable even for people that doesn't know these specific libraries or implementation internals. After all this proxy is the prime example for how to use `imap-flow`.

proxy/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0.75"
+argh = "0.1.12"
```
jakoschiko, line 8: I didn't know `argh` before, but it looks okay.

proxy/.gitignore
duesee: > Doesn't git allow a global .gitignore in the home directory for such use cases? Otherwise you would need to add personal stuff that is not related to the project (IDE config, etc.) to all local .gitignore files.

Yes, it does. I kinda prefer local `.gitignores` when the folder at hand is something that could as well make sense on it's own. It also keeps effects local. Maybe it's a matter of taste. Do you prefer to change it?

> Why not? It's convenient. IMO convenience is more important than cleanness here.

I agree that it's convinient for us :-) But it could also raise some questions. Why is it GMX and not provider X? Is this project affiliated with Mailbox.org? I would prefer to not have existing providers it in the default config.

proxy/src/main.rs
duesee: Okay! Let me reintroduce it.

proxy/.gitignore
jakoschiko: > Do you prefer to change it?

It doesn't bother me (maybe later when I create a rust module named `private` and will be confused...). I just thought that you are spending too much time managing and explaining me your `.gitignore` file.

> I agree that it's convinient for us :-) But it could raise some questions. Why is it GMX and not provider X? Is this project affiliated with Mailbox.org? Thus, I would prefer to not have existing providers in the default config.

I'd prefer to have them all :p

duesee: > I'd prefer to have them all :p

Any idea how to resolve this dilemma? 🙈 

I would be okay with an additional comment in the config. But "These are here because the devs use them for testing" is kinda... not right.

jakoschiko: I don't mind enough to resist any longer, so feel free to ignore my comment.

And I'm sure we are not the only devs that are using these free providers for testing :p

PR #56, opened 2023-10-25
feat: Improve handle handling

### LINKED ISSUE #45: Should we handle `Status::No` in `ClientFlow::maybe_abort_command`?

jakoschiko:
Before this PR:

- If `ClientFlow` receives a `Status::Bad` with the matching tag then the current `Command` will be aborted
- All received `Status` (including the one mentioned before) will be emitted via `ClientFlowEvent::StatusReceived`
  - `ClientFlowEvent::StatusReceived` does not contain the `ClientFlowCommandHandle`
  - The user of `ClientFlow` must do complicated logic in order to detect the abortion of the `Command`

After this PR:

- If `ClientFlow` receives a `Status::Bad` with the matching tag then the current `Command` will be aborted and `ClientFlowEvent::CommandRejected` will be emitted
  -  `ClientFlowEvent::CommandRejected` contains the `ClientFlowCommandHandle`
- All other received `Status` will be emitted via `ClientFlowEvent::StatusReceived`

Why this might be a good idea:

- Sending the `Command` (and handling literals) is the responsibility of `imap-flow`
- In order to do so `ClientFlow` needs to handle `Status::Bad` in some cases
- By introducing a new variant for `ClientFlowEvent` it's now clear that:
  - in case of `ClientFlowEvent::CommandRejected` the `Status` was already handled by `ClientFlow`
  - in case of `ClientFlowEvent::StatusReceived` the `Status` needs to be handled by the user of `ClientFlow`
    -  the user of `ClientFlow` can more easily detect an unexpected `Status` and log it
- The user of `ClientFlow` can simply match on `ClientFlowEvent::CommandSent` or `ClientFlowEvent::CommandRejected` to detect when the sending process of the `Command` is finished

I also improved the documentation for both `ClientFlowCommandHandle` and `ServerFlowResponseHandle`.

Closes #45

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -26,10 +26,18 @@ async fn main() {
     loop {
         match client.progress().await.unwrap() {
             ClientFlowEvent::CommandSent {
-                tag,
                 handle: got_handle,
+                tag,
             } => {
-                println!("command sent: {tag:?}, {handle:?}");
+                println!("command sent: {got_handle:?}, {tag:?}");
+                assert_eq!(handle, got_handle);
+            }
+            ClientFlowEvent::CommandRejected {
+                handle: got_handle,
+                tag,
+                status,
+            } => {
+                println!("command rejected: {got_handle:?}, {tag:?}, {status:?}");
                 assert_eq!(handle, got_handle);
             }
             ClientFlowEvent::DataReceived { data } => {
--- a/proxy/src/proxy.rs
+++ b/proxy/src/proxy.rs
@@ -1,6 +1,10 @@
 use std::{net::SocketAddr, sync::Arc};
 
 use colored::Colorize;
+use imap_codec::imap_types::{
+    core::Text,
+    response::{Code, Status},
+};
 use imap_flow::{
     client::{ClientFlow, ClientFlowError, ClientFlowEvent, ClientFlowOptions},
     server::{ServerFlow, ServerFlowError, ServerFlowEvent, ServerFlowOptions},
@@ -18,6 +22,10 @@ use crate::{
     config::Service,
     util::{self, ControlFlow},
 };
+
+const LITERAL_ACCEPT_TEXT: &str = "proxy: Literal accepted by proxy";
+const LITERAL_REJECT_TEXT: &str = "proxy: Literal rejected by proxy";
+const COMMAND_REJECTED_TEXT: &str = "proxy: Command rejected by server";
 
 #[derive(Debug, Error)]
 pub enum ProxyError {
@@ -165,7 +173,11 @@ impl Proxy<ConnectedState> {
 
         let mut client_to_proxy = {
             // TODO: Read options from config
-            let options = ServerFlowOptions::default();
+            let options = ServerFlowOptions {
+                literal_accept_text: Text::try_from(LITERAL_ACCEPT_TEXT).unwrap(),
+                literal_reject_text: Text::try_from(LITERAL_REJECT_TEXT).unwrap(),
+                ..Default::default()
+            };
 
             let result =
                 ServerFlow::send_greeting(self.state.client_to_proxy, options, greeting).await;
@@ -263,11 +275,39 @@ fn handle_server_event(
 
     match event {
         ClientFlowEvent::CommandSent {
-            tag,
             handle: _handle,
+            tag,
         } => {
             // TODO: log handle
             trace!(role = "p2s", ?tag, "---> Forwarded command");
+        }
+        ClientFlowEvent::CommandRejected {
+            handle: _handle,
+            tag,
+            status,
+        } => {
+            // TODO: log handle
+            trace!(role = "p2s", ?tag, ?status, "---> Aborted command");
+            let status = match status.code() {
+                Some(Code::Alert) => {
+                    // Keep the alert message because it MUST be displayed to the user
+                    Status::Bad {
+                        tag: Some(tag),
+                        code: Some(Code::Alert),
+                        text: status.text().clone(),
+                    }
+                }
+                _ => {
+                    // Use generic message because the original code and text might be misleading
+                    Status::Bad {
+                        tag: Some(tag),
+                        code: None,
+                        text: Text::try_from(COMMAND_REJECTED_TEXT).unwrap(),
+                    }
+                }
+            };
+            let _handle = client_to_proxy.enqueue_status(status);
+            // TODO: log handle
         }
         ClientFlowEvent::DataReceived { mut data } => {
             trace!(data=%format!("{:?}", data).blue(), role = "s2p", "<--| Received data");
--- a/src/client.rs
+++ b/src/client.rs
@@ -82,6 +82,11 @@ impl ClientFlow {
         Ok((client_flow, greeting))
     }
 
+    /// Enqueues the [`Command`] for being sent to the client.
+    ///
+    /// The [`Command`] is not sent immediately but during one of the next calls of
+    /// [`ClientFlow::progress`]. All [`Command`]s are sent in the same order they have been
+    /// enqueued.
     pub fn enqueue_command(&mut self, command: Command<'_>) -> ClientFlowCommandHandle {
         let handle = self.next_command_handle;
         self.next_command_handle = ClientFlowCommandHandle(handle.0 + 1);
@@ -104,7 +109,7 @@ impl ClientFlow {
 
     async fn progress_command(&mut self) -> Result<Option<ClientFlowEvent>, ClientFlowError> {
         match self.send_command_state.progress(&mut self.stream).await? {
-            Some((tag, handle)) => Ok(Some(ClientFlowEvent::CommandSent { tag, handle })),
+            Some((tag, handle)) => Ok(Some(ClientFlowEvent::CommandSent { handle, tag })),
             None => Ok(None),
         }
     }
@@ -139,8 +144,15 @@ impl ClientFlow {
 
             match response {
                 Response::Status(status) => {
-                    self.maybe_abort_command(&status);
-                    break Some(ClientFlowEvent::StatusReceived { status });
+                    if let Some((tag, handle)) = self.maybe_abort_command(&status) {
+                        break Some(ClientFlowEvent::CommandRejected {
+                            handle,
+                            tag,
+                            status,
+                        });
+                    } else {
+                        break Some(ClientFlowEvent::StatusReceived { status });
+                    }
                 }
                 Response::Data(data) => break Some(ClientFlowEvent::DataReceived { data }),
                 Response::CommandContinuationRequest(_) => {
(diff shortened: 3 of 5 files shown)


proxy/src/proxy.rs
```diff
@@ -263,12 +263,20 @@ fn handle_server_event(
...
             // TODO: log handle
             trace!(role = "p2s", ?tag, "---> Forwarded command");
         }
+        ClientFlowEvent::CommandRejected {
+            handle: _handle,
+            tag,
+            status,
+        } => {
+            // TODO: log handle
+            trace!(role = "p2s", ?tag, ?status, "---> Aborted command");
```
jakoschiko, line 278: Do we need to forward the status here? It's complicated, think twice!

duesee: I think "yes".

But we have an interesting problem here (that I'm sure we discussed at some point but denied it :-))

Situation:

```
// client <--> proxy
C: A1 LOGIN {999...999}
P: + ... // Our proxy never rejects literals on it's own
C: <username> "password"

// proxy <--> server
P: A1 LOGIN {999...999} // forwarded
S: A1 BAD // Server is not happy (but our proxy already accepted!)
```

So, the question is now: How do we make sure to not change the semantics of the connection. And I think: we can't fully. IF we want to have a complete command (in order to have all flexibility to change every aspect of it), we cannot forward "half" commands and need to do the literal dance on each side.

But, luckily, I think, the client should not care too much about *why* a command was rejected. We can also reject it after we received all literals.

Meaning: I think `BAD` should be forwarded as it was before.

Additional:

* I think we should also filter capabilities here so that it is fundamentally not possible to exchange unsupported capabilities in any way with the proxy.

duesee (commented):
Thanks! I left some suggestions that could be applied immediately. I'm not already accepting the PR as we probably want to discuss a few things inline. Happy to learn if thinking twice helped :D

proxy/src/proxy.rs
duesee: Additional:

* I think we should also filter capabilities here so that it is fundamentally not possible to exchange unsupported capabilities in any way with the proxy.

src/client.rs
```diff
@@ -82,6 +82,11 @@ impl ClientFlow {
         Ok((client_flow, greeting))
     }
 
+    /// Enqueues the [`Command`] for being sent to the client.
+    ///
+    /// The [`Command`] is not sent immediately but during one of the next calls of
+    /// [`ClientFlow::progress`]. All [`Command`]s are sent in the same order they has been
```
duesee, line 88:
```suggestion
    /// [`ClientFlow::progress`]. All [`Command`]s are sent in the same order they have been
```

src/client.rs
```diff
@@ -153,33 +165,49 @@ impl ClientFlow {
...
         tag: Tag<'static>,
+    },
+    /// The enqueued [`Command`] wasn't sent completely because the server rejected it.
+    CommandRejected {
+        /// The handle of the enqueued [`Command`].
         handle: ClientFlowCommandHandle,
+        /// The [`Tag`] of the enqueued [`Command`].
+        tag: Tag<'static>,
+        /// The [`Status`] sent by the server in order to reject the [`Command`].
+        /// This [`Status`] was already handled by [`ClientFlow`] and can safely be ignored.
```
duesee, line 209: It could still have an `Code::Alert` telling the client to display an error message. Maybe we should say that the "flow", e.g., aborting the command was handled but the code and text could still be used.

src/server.rs
```diff
@@ -84,13 +84,23 @@ impl ServerFlow {
         Ok(server_flow)
     }
 
+    /// Enqueues the [`Data`] response for being sent to the client.
+    ///
+    /// The response is not sent immediately but during one of the next calls of
+    /// [`ServerFlow::progress`]. All responses are sent in the same order they has been
```
duesee, line 90:
```suggestion
    /// [`ServerFlow::progress`]. All responses are sent in the same order they have been
```

src/server.rs
```diff
@@ -84,13 +84,23 @@ impl ServerFlow {
...
         let handle = self.next_response_handle();
         self.send_response_state
             .enqueue(Some(handle), Response::Data(data));
         handle
     }
 
+    /// Enqueues the [`Status`] response for being sent to the client.
+    ///
+    /// The response is not sent immediately but during one of the next calls of
+    /// [`ServerFlow::progress`]. All responses are sent in the same order they has been
```
duesee, line 102:
```suggestion
    /// [`ServerFlow::progress`]. All responses are sent in the same order they have been
```

duesee (changes requested):
Let's talk at some point about our PR process. I'll use "Request changes" to signal that I believe we could improve the PR. (Even when the improvement is minor and can be applied fast.) Feel free to re-request a review at any time! I'll also re-request reviews from you until I get an "Accept" :-)

I'll also try to always say how important a requested change is for me. When you feel it's meant as "optional" from me and also don't think you want to handle it just now, just click on "resolve" and close the discussion.

I think we should try to merge PRs as fast as possible when they improve the status quo and don't have meaningful disadvantages.

src/client.rs
```diff
@@ -139,8 +144,15 @@ impl ClientFlow {
 
             match response {
                 Response::Status(status) => {
-                    self.maybe_abort_command(&status);
-                    break Some(ClientFlowEvent::StatusReceived { status });
+                    if let Some((tag, handle)) = self.maybe_abort_command(&status) {
+                        break Some(ClientFlowEvent::CommandRejected {
+                            handle,
+                            tag,
+                            status,
```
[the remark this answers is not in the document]
jakoschiko: Some arguments for keeping it:
- It probably doesn't matter, the user will ignore it or just log it
- If at some point in time we notice that there are serious servers that are seriously using `::No` for rejecting literals, then we can change `ClientFlow` so that it matches and returns `::No` without changing the API

I'd prefer to keep it and to revisit it later.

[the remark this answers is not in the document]
duesee: Good point 👍🏻

proxy/src/proxy.rs
jakoschiko: Options:

1. Don't send any `Status`:
    - Probably bad idea because it might cause a dead lock in the real client
2. Forward the given `Status::Bad`
    - Is a `::Bad` semantically always okay? Does it contain some text or similar things that might be misleading?
3. Don't send the given `Status::Bad`, but send another `Status` instead
    - Maybe `::No`? Maybe change some meta fields inside the `::Bad`?

Let's wait until Monday for a decision.

examples/client.rs
```diff
@@ -26,10 +26,18 @@ async fn main() {
...
+                println!("command sent: {got_handle:?}, {tag:?}");
+                assert_eq!(handle, got_handle);
+            }
+            ClientFlowEvent::CommandRejected {
                 handle: got_handle,
+                tag,
+                status,
             } => {
-                println!("command sent: {tag:?}, {handle:?}");
+                println!("command aborted: {got_handle:?}, {tag:?}, {status:?}");
```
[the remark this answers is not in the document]
jakoschiko: Previously I called the event `ClientFlowEven::CommandAborted` because it makes it clear that `Command` wasn't sent completely. But then I renamed it to `ClientFlowEvent::CommandRejected` because the RFC is using the term "reject". What do you think is the better name?

src/client.rs
jakoschiko: I didn't write that it's forbidden to use it. But still, I'll try to find a better phrase.

examples/client.rs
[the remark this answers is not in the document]
duesee: I prefer "reject" because the server rejecting the command is the reason we need to abort sending it. It somehow captures "more" of what happend.

src/client.rs
duesee: Thanks :-)

proxy/src/proxy.rs
duesee: > 1. Don't send any Status:
> Probably bad idea because it might cause a dead lock in the real client

Yeah :-/ I played around a bit and this is exactly what happens.

> 2. Forward the given Status::Bad
> Is a ::Bad semantically always okay? Does it contain some text or similar things that might be misleading?

Unfortunately, also yeah :-/

When we forward the `Bad` the client could see the following (confusing) trace:

```imap
C: A1 LOGIN {999...999}
P: + ...
C: <username> "password"
S: A1 BAD literal rejected
```

As far as I know, the `text` is the only thing that can make this confusing. But, still, I think this is as close as we can get w/o loosing flexibility.

Worth to mention: From the IMAP RFC:

> A rejected command (BAD response) never changes the state of the connection or of the selected mailbox. A failed command (NO response) generally does not change the state of the connection or of the selected mailbox; the exception being the SELECT and EXAMINE commands.

> 3. Don't send the given Status::Bad, but send another Status instead
> Maybe ::No? Maybe change some meta fields inside the ::Bad?

I'm not sure, really. Hard to overlook the implications. Also given that `No` can potentially change state and `Bad` never does (if we can trust the RFC).

jakoschiko: Then I'd vote for `::Bad` with a custom text.

duesee: Oh, good idea :+1:

proxy/src/proxy.rs
```diff
@@ -19,6 +23,10 @@ use crate::{
     util::{self, ControlFlow},
 };
 
+const LITERAL_ACCEPT_TEXT: &str = "proxy: Literal accepted by proxy";
+const LITERAL_REJECT_TEXT: &str = "proxy: Literal rejected by proxy";
+const COMMAND_REJECTED_TEXT: &str = "proxy: Command rejected by server";
```
jakoschiko, line 28: The actual text is up for discussion. But messages like these might help to identify IMAP messages that are not forwarded but instead generated by the proxy itself.

proxy/src/proxy.rs
jakoschiko: I added complexity. Is it good now?

proxy/src/proxy.rs
```diff
@@ -263,12 +275,40 @@ fn handle_server_event(
...
+            status,
+        } => {
+            // TODO: log handle
+            trace!(role = "p2s", ?tag, ?status, "---> Aborted command");
+            let status = match status.code() {
+                Some(Code::Alert) => {
+                    // Keep the alert message because it MUST be displayed to the user
+                    Status::Bad {
+                        tag: Some(tag),
+                        code: Code::Alert,
```
duesee, line 296: I think it must be ...

```suggestion
                        code: Some(Code::Alert),
```

proxy/src/proxy.rs
duesee: Hehe, I hate that we have to care, but it's beautiful how easy we can make those changes! :-)

I'm not sure if it's a good or bad idea to filter the `code`. My intuition would be to keep changes to a minimum, i.e., remove unsupported capabilities if the `code` is `Code::Capability`, and add a `proxy: please refer to ... ` prefix to the original `text`.

`Alert` is certainly a thing to keep in mind. But are there more codes? I don't know. That's why I feel we should rather forward what we don't know.

duesee (approved):
Thanks! I'm already approving. Feel free to make smaller changes, but this is good to go! (I commited a small fix to make CI work, maybe you want to squash.)

proxy/src/proxy.rs
duesee: Resolving as discussed privately.

PR #60, opened 2023-11-04
feat: First draft of `tasks`

duesee:
This is the current state of the `tasks` crate. I already like how working with it feels, but I am relatively sure we will want to change a few things as we go.

The next thing I would like to do is to port the "CLI demo" -- the thing that fetches all status data items from a server -- from the `duesee_cli` branch to `tasks` and fix everything on the way.

There are some open questions regarding error handling (that we probably should discuss in person), an `unwrap`, and a few missed opportunities for better code.

I feel that we can still be more liberal with anything that is not `imap-flow` for now. Thus, I'm primarily making this PR to have something usable online. What do you think?

By the way, I kept all commits in case they're helpful, but we probably should squash to a single commit listing us both as authors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,4 +19,8 @@ resolver = "2"
 members = [
     "proxy",
     "tag-generator",
+    "tasks",
 ]
+
+[patch.crates-io]
+imap-types = { git = "https://github.com/duesee/imap-codec" }
--- a/tasks/src/lib.rs
+++ b/tasks/src/lib.rs
@@ -1,0 +1,332 @@
+use std::{
+    any::Any,
+    collections::HashMap,
+    fmt::{Debug, Formatter},
+    marker::PhantomData,
+};
+
+use imap_codec::imap_types::{
+    command::{Command, CommandBody},
+    core::Tag,
+    response::{Bye, Data, Response, Status, StatusBody, StatusV2, Tagged},
+};
+use imap_flow::client::{ClientFlow, ClientFlowCommandHandle, ClientFlowError, ClientFlowEvent};
+use tag_generator::TagGenerator;
+use thiserror::Error;
+
+pub mod tasks;
+
+/// Tells how a specific IMAP [`Command`] is processed.
+///
+/// Most `process_` trait methods consume interesting responses (returning `None`),
+/// and move out uninteresting responses (returning `Some(...)`).
+///
+/// If no active task is interested in a given response, we call this response "unsolicited".
+pub trait Task: 'static {
+    /// Output of the task.
+    ///
+    /// Returned in [`Self::process_tagged`].
+    type Output;
+
+    /// Returns the [`CommandBody`] to issue for this task.
+    ///
+    /// Note: The [`Scheduler`] will tag the [`CommandBody`] creating a complete [`Command`].
+    fn command_body(&self) -> CommandBody<'static>;
+
+    /// Process data response.
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        // Default: Don't process server data
+        Some(data)
+    }
+
+    /// Process an untagged response.
+    fn process_untagged(
+        &mut self,
+        status_body: StatusBody<'static>,
+    ) -> Option<StatusBody<'static>> {
+        // Default: Don't process untagged status
+        Some(status_body)
+    }
+
+    /// Process bye response.
+    fn process_bye(&mut self, bye: Bye<'static>) -> Option<Bye<'static>> {
+        // Default: Don't process bye
+        Some(bye)
+    }
+
+    /// Process command completion result response.
+    ///
+    /// The [`Scheduler`] already chooses the corresponding response by tag.
+    fn process_tagged(self, status_body: StatusBody<'static>) -> Self::Output;
+}
+
+/// Scheduler managing enqueued tasks and routing incoming responses to active tasks.
+pub struct Scheduler {
+    flow: ClientFlow,
+    waiting_tasks: HashMap<ClientFlowCommandHandle, Box<dyn TaskAny>>,
+    active_tasks: HashMap<ClientFlowCommandHandle, (Tag<'static>, Box<dyn TaskAny>)>,
+    tag_generator: TagGenerator,
+}
+
+impl Scheduler {
+    /// Create a new scheduler.
+    pub fn new(flow: ClientFlow) -> Self {
+        Self {
+            flow,
+            waiting_tasks: Default::default(),
+            active_tasks: Default::default(),
+            tag_generator: TagGenerator::new(),
+        }
+    }
+
+    /// Enqueue a [`Task`].
+    pub fn enqueue_task<T>(&mut self, task: T) -> TaskHandle<T>
+    where
+        T: Task,
+    {
+        let cmd = {
+            let tag = self.tag_generator.generate();
+            let body = task.command_body();
+
+            Command { tag, body }
+        };
+
+        let handle = self.flow.enqueue_command(cmd);
+
+        let replaced = self.waiting_tasks.insert(handle, Box::new(task));
+        assert!(replaced.is_none());
+
+        TaskHandle::new(handle)
+    }
+
+    /// Progress the connection returning the next event.
+    pub async fn progress(&mut self) -> Result<SchedulerEvent, SchedulerError> {
+        loop {
+            let event = self.flow.progress().await?;
+
+            match event {
+                ClientFlowEvent::CommandSent { tag, handle } => {
+                    // This `unwrap` can't fail because `waiting_tasks` contains all unsent `Commands`.
+                    let entry = self.waiting_tasks.remove(&handle).unwrap();
+                    self.active_tasks.insert(handle, (tag, entry));
+                }
+                ClientFlowEvent::CommandRejected { handle, status, .. } => {
+                    let body = match StatusV2::from(status) {
+                        StatusV2::Tagged(Tagged { body, .. }) => body,
+                        _ => unreachable!(),
+                    };
+
+                    // This `unwrap` can't fail because `active_tasks` contains all in-progress `Commands`.
+                    let (_, task) = self.active_tasks.remove(&handle).unwrap();
+
+                    let output = Some(task.process_tagged(body));
+
+                    return Ok(SchedulerEvent::TaskFinished(TaskToken { handle, output }));
+                }
+                ClientFlowEvent::DataReceived { data } => {
+                    if let Some(data) = trickle_down(data, self.active_tasks_mut(), |task, data| {
+                        task.process_data(data)
+                    }) {
+                        return Ok(SchedulerEvent::Unsolicited(Response::Data(data)));
+                    }
+                }
+                ClientFlowEvent::StatusReceived { status } => match StatusV2::from(status) {
+                    StatusV2::Untagged(body) => {
+                        if let Some(body) =
+                            trickle_down(body, self.active_tasks_mut(), |task, body| {
+                                task.process_untagged(body)
+                            })
+                        {
+                            return Ok(SchedulerEvent::Unsolicited(Response::Status(
+                                Status::from(StatusV2::Untagged(body)),
+                            )));
+                        }
+                    }
+                    StatusV2::Bye(bye) => {
+                        if let Some(bye) =
+                            trickle_down(bye, self.active_tasks_mut(), |task, bye| {
+                                task.process_bye(bye)
+                            })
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -19,4 +19,8 @@ resolver = "2"
 members = [
     "proxy",
     "tag-generator",
+    "tasks",
 ]
+
+[patch.crates-io]
+imap-types = { git = "https://github.com/duesee/imap-codec" }
```
duesee, line 26: This makes it simpler for now to work in lock-step with `imap-codec`. Should I file an issue as a reminder to remove this?

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+        let replaced = self.waiting_tasks.insert(handle, Box::new(task));
+        assert!(replaced.is_none());
+
+        TaskHandle::new(handle)
+    }
+
+    pub async fn progress(&mut self) -> SchedulerEvent {
+        loop {
+            // TODO: How to handle error?
+            let event = self.flow.progress().await.unwrap();
```
duesee, line 102: The mentioned `unwrap`.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+                        let (handle, _) = self
+                            .active_tasks
+                            .iter()
+                            .find(|(_, (tag_, _))| tag == *tag_)
+                            .unwrap();
+
+                        *handle
+                    };
+
+                    let body = match StatusV2::from(status) {
```
duesee, line 121: I commited `StatusV2` to imap-codec. `Status` and `StatusV2` are isomorph -- tight types, yay! -- but V2 is much more pleasent to work with. Not sure how to continue with this...

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    _phantom: PhantomData<T>,
+}
+
+impl<T: Task> Debug for TaskHandle<T> {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        f.debug_struct("TaskHandle")
+            .field("handle", &self.handle)
+            .finish()
+    }
+}
```
jakoschiko, line 208: fyi: What we want here but Rust doesn't support is called perfect derive, e.g. see https://crates.io/crates/perfect-derive or the linked blog post https://smallcultfollowing.com/babysteps//blog/2022/04/12/implied-bounds-and-perfect-derive.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+impl<T: Task> PartialEq for TaskHandle<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.handle == other.handle
+    }
+}
+
+impl<T: Task> TaskHandle<T>
+where
+    T: 'static,
```
jakoschiko, line 226: `'static` shouldn't  be necessary anymore because `Task` has this bound already.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    fn new(handle: ClientFlowCommandHandle) -> Self {
+        Self {
+            handle,
+            _phantom: Default::default(),
+        }
+    }
+
+    /// Try resolving the task invalidating the token.
+    ///
+    /// The token is invalidated iff the return value is `Some()`.
```
jakoschiko, line 237: I prefer `Some` instead of this pseudo-syntax.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    pub fn resolve(&self, token: &mut TaskToken) -> Option<T::Output> {
+        if token.handle != self.handle {
+            return None;
+        }
+
+        let output = token.output.take()?;
+        let output = output.downcast::<T::Output>().unwrap();
```
jakoschiko, line 244: It's like magic.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+use imap_codec::imap_types::{
+    command::{Command, CommandBody},
+    core::Tag,
+    response::{Bye, Data, Response, Status, StatusBody, StatusV2, Tagged},
+};
+use imap_flow::client::{ClientFlow, ClientFlowCommandHandle, ClientFlowEvent};
+use tag_generator::TagGenerator;
+
+pub mod tasks;
```
jakoschiko, line 16: What's better?
```
use
mod
code
```
Or?
```
mod
use
code
```
IMO the second one because both `use` and `code` might depend on `mod` but not vice versa.

tasks/src/tasks.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::Task;
+
+#[derive(Default)]
+pub struct CapabilityTask {
+    /// We use this as scratch space.
+    capabilities: Option<NonEmptyVec<Capability<'static>>>,
+}
+
+impl Task for CapabilityTask {
+    type Output = Result<NonEmptyVec<Capability<'static>>, &'static str>;
```
jakoschiko, line 20: Should `Result` be part of `Task`? I.e. `process_tagged` has the return type `Result<Self::Output, Self::Error>`.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    fn active_tasks_mut(&mut self) -> impl Iterator<Item = &mut Box<dyn TaskAny>> {
+        self.active_tasks.values_mut().map(|(_, task)| task)
+    }
+}
+
+#[derive(Debug)]
+pub enum SchedulerEvent {
+    TaskFinished(TaskToken),
+    Shutdown,
```
jakoschiko, line 192: What's this?

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+                            )));
+                        }
+                    }
+                    StatusV2::Tagged(Tagged { tag, body }) => {
+                        let handle = {
+                            let (handle, _) = self
+                                .active_tasks
+                                .iter()
+                                .find(|(_, (tag_, _))| tag == *tag_)
+                                .unwrap();
```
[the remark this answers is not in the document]
duesee: Oh, good catch!

`Unsolicited` would work, but ...

Do we ever give up on a connection?
Like... what needs to happen that we stop a session before we cause damage?

Getting a tagged response to a command we never send is pretty fatal, IMO, as all bets are off.

tasks/src/lib.rs
duesee: Btw., did you know you can use a "suggestion" block in GitHub like ...

```suggestion
    /// The token is invalidated iff the return value is `Some`.
```

It's a bit faster than to explain simple changes, and a PR author can automatically accept all suggestions in a batch.

tasks/src/lib.rs
duesee: Thanks for the link! I stumbled upon this in another project and learned that there are already crates that do what perfect derive would do. (Was it https://github.com/mcarton/rust-derivative?) But we needed a custom derive that wasn't supported as far as I remember. Should this come up a few times, we could evaluate if we want to use a helper crate.

tasks/src/tasks.rs
duesee: I'm not sure. I would propose we implement a few more `Task`s to gain a better understanding first.

tasks/src/lib.rs
duesee: I like the second one, because having something like ...

```rust
use ...;

mod foo;
use foo::bar;
```

is very cohesive (and visually documenting.)

But I think I started to use the first variant due to auto-import and rustfmt ...

```rust
use ...;

mod foo;
use foo::bar;
// newline deleted by rustfmt?
mod baz;
use baz::bar;
use std::io::Write; // Auto-import. Kept by rustfmt?
```

We can try again and see how it goes.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+        Some(bye)
+    }
+
+    /// Process command completion result response.
+    ///
+    /// The [`Scheduler`] already chooses the corresponding response by tag.
+    fn process_tagged(self, status_body: StatusBody<'static>) -> Self::Output;
+}
+
+/// Scheduler.
```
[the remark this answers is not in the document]
duesee: Bad 🐶

tasks/src/lib.rs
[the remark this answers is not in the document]
jakoschiko: But we don't want to crash the client only because the server is trash. Let the caller decide, we can't give up the connection here. The only thing we can do here is emitting an event with a more seriously sounding name than `Unsolicited`.

tasks/src/lib.rs
jakoschiko: Yeah, I'm just lazy :p

tasks/src/lib.rs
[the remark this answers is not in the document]
duesee: Oh, you misunderstood. I didn't propose to panic. I never will, in fact :D I was just thinking if we should return, e.g., the `Event::Shutdown` and "fuse" the flow to make it unusable.

tasks/src/lib.rs
jakoschiko: I meant that all `mod`s were always at the top before any `use`s. But I don't know what's work best with `cargo fmt`.

tasks/src/lib.rs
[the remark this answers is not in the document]
jakoschiko: Let the caller decide :p

But be opinionated by returning a `FatalUnknownTagStatusFromCthulhu` event.

tasks/src/lib.rs
duesee: Should work. Let's do it if you prefer.

duesee: I feel you.

tasks/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+                ClientFlowEvent::CommandRejected { tag, status, .. } => {
+                    let handle = {
+                        let (handle, _) = self
+                            .active_tasks
+                            .iter()
+                            .find(|(_, (tag_, _))| tag == *tag_)
+                            .unwrap();
+
+                        *handle
+                    };
```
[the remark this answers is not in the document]
duesee: Leftover from the old days.

tasks/src/lib.rs
duesee: Let's make this a separate issue and do all changes at once when we feel like it.

tasks/src/lib.rs
duesee: The thing I could use instead of `FatalUnknownTagStatusFromCthulhu`? :-)

duesee: I did something different. The reason for the `Shutdown` variant was that an IMAP server can send a `bye` at any time. A client receving an unsolicited bye is then expected to read as many responses as it can (until the server closes the connection). This is an expected outcome and not an error. Without `Shutdown` we basically assume a connection lives forever. Still: Let's postpone this discussion. I'll remove `Shutdown` for now.

duesee: Forget what I said... delted. Shutdown is, in fact, `Unsolicited(Bye)?. There should be no need to signal it separately.

tasks/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+                    }
+                    StatusV2::Tagged(Tagged { tag, body }) => {
+                        let handle = {
+                            if let Some((handle, _)) =
+                                self.active_tasks.iter().find(|(_, (tag_, _))| tag == *tag_)
+                            {
+                                *handle
+                            } else {
+                                return Err(SchedulerError::FatalProtocolIssue(
+                                    FatalProtocolIssue::UnexpectedTaggedResponse,
```
jakoschiko, line 163: `UnexpectedTaggedResponse` should contain the response. Then the caller is able to log it.

tasks/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// Flow error
+    #[error("flow error")]
+    Flow(#[from] ClientFlowError),
+    /// Fatal protocol issue
+    #[error("fatal protocol issue")]
+    FatalProtocolIssue(FatalProtocolIssue),
+}
+
+#[derive(Debug, Error)]
+pub enum FatalProtocolIssue {
```
jakoschiko, line 201: Do we really need this enum? Isn't `ClientFlowError` also a fatal protocol issue?

duesee: Not sure... but let's err on the side of less code :+1:

jakoschiko (approved):

PR #71, opened 2023-11-14
feat: Implement AUTHENTICATE flow

### LINKED ISSUE #16: feat: Implement AUTHENTICATE flow

duesee:
Closes #16 

```imap
S: * OK Hello!
C: A1 CAPABILITY
S: * capability imap4rev1 auth=plain
S: * OK [ALERT] Booooring
S: A1 OK fine
C: A2 AUTHENTICATE PLAIN
S: * 1337 exists
S: * NO [ALERT] Are you serious?
S: * 1 expunge
C: + FINE!!!
     ^
     |
     Base64/Text?
C: AGFsaWNlAHBhc3N3b3Jk
S: A2 NO :P
```

```rust
received greeting: Greeting { kind: Ok, code: None, text: Text("Hello!") }
command sent: ClientFlowCommandHandle(0), Tag("A1")
data received: Capability([Imap4Rev1, Auth(Plain)]+)
status received: Untagged(StatusBody { kind: Ok, code: Some(Alert), text: Text("Booooring") })
command completion result response received: Tagged(Tagged { tag: Tag("A1"), body: StatusBody { kind: Ok, code: None, text: Text("fine") } })
data received: Exists(1337)
status received: Untagged(StatusBody { kind: No, code: Some(Alert), text: Text("Are you serious?") })
data received: Expunge(1)
command sent: ClientFlowCommandHandle(1), Tag("A2")
command completion result response received: Tagged(Tagged { tag: Tag("A2"), body: StatusBody { kind: No, code: None, text: Text(":P") } })
```

# TODO

* [ ] Proxy
* [ ] Base64 vs. Text
* [ ] Generic SASL

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -1,8 +1,11 @@
 use imap_codec::imap_types::{
     command::{Command, CommandBody},
     core::Tag,
+    response::Status,
+    secret::Secret,
 };
 use imap_flow::{
+    auth::Authenticate,
     client::{ClientFlow, ClientFlowEvent, ClientFlowOptions},
     stream::AnyStream,
 };
@@ -18,9 +21,10 @@ async fn main() {
             .unwrap();
     println!("received greeting: {greeting:?}");
 
-    let handle = client.enqueue_command(Command {
+    // TODO: Forbid Command::Authenticate.
+    let handle1 = client.enqueue_command(Command {
         tag: Tag::try_from("A1").unwrap(),
-        body: CommandBody::Noop,
+        body: CommandBody::Capability,
     });
 
     loop {
@@ -30,7 +34,7 @@ async fn main() {
                 tag,
             } => {
                 println!("command sent: {got_handle:?}, {tag:?}");
-                assert_eq!(handle, got_handle);
+                assert_eq!(handle1, got_handle);
             }
             ClientFlowEvent::CommandRejected {
                 handle: got_handle,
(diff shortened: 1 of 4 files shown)


duesee:
Boooh, this won't work generically for all SASL mechanisms. Let's orientate on this https://datatracker.ietf.org/doc/html/rfc3501#section-6.2.2. If we can make GSSAPI work, everything should. (It's multi-stage challenge/response.)

dequbed:
> Boooh, this won't work generically for all SASL mechanisms. Let's orientate on this https://datatracker.ietf.org/doc/html/rfc3501#section-6.2.2. If we can make GSSAPI work, everything should. (It's multi-stage challenge/response.)

Mind you that RFC9051 (which obsoletes RFC3501) extends `AUTHENTICATE` to also be able to send an initial response. So your server side will have to handle being sent e.g. `A2 AUTHENTICATE PLAIN AGZvbwBiYXI=`.

dequbed:
> Cool that you write here! :-)
> 
> The problem I'm trying to solve is abstracting over IMAP's authentication flow so that users don't have to care too much. `imap-flow` abstracts over literal handling, so command continuation request are basically hidden. (Which is good IMO!)
> 
> But... for AUTHENTICATE, this is different: I think we _need_ to forward every "SASL step". Otherwise, we can't [proxy](https://github.com/duesee/imap-flow/tree/main/proxy) authentication. With this in mind, your `rsasl` library could fit _perfectly_ 🙌

There's a snag you might hit there in that authentication mechanisms supporting Channel Binding (`SCRAM`, the `GS2-*` family, the `OPAQUE` mechanism I'm working on right now, and a few others) can not be proxied, by design. Unless you have a cooperating server or you can terminate authentication in the proxy and then do a completely independent authentication with the server proxying authentication with one of those mechanisms will always fail. Granted, those mechanisms are quite rare in the E-Mail space so this is probably not a relevant issue in practice, but at least Dovecot does actually support SCRAM with channel binding (the `-PLUS` variant).

duesee:
> Unless you have a cooperating server or you can terminate authentication in the proxy and then do a completely independent authentication with the server proxying authentication with one of those mechanisms will always fail.

That's a great use case for the proxy! Terminate old authentication mechanisms and upgrade to the fancy ones with channel binding :-) Interesting to hear that Dovecot supports `SCRAM-*-PLUS`.

> There's a snag you might hit there in that authentication mechanisms using Channel Binding (SCRAM-*-PLUS, the GS2-* family, the OPAQUE mechanism I'm working on right now, and a few others) can not be proxied, by design.

This is definitely something to keep in mind, thanks! Currently, we strip all authentication methods in the proxy. Later, we could terminate (with the possibility to upgrade), forward "step-by-step", or reject them.

PR #73, opened 2023-11-14
chore: Use `imap-{types,codec}` `v2`

duesee:
Only chore to make the review of #71 easier. I don't think there is much to review here. I'll go ahead and merge. Hope that's okay :-)

### DIFF
--- a/src/send.rs
+++ b/src/send.rs
@@ -122,6 +122,9 @@ impl<K> SendCommandState<K> {
                         });
                         break true;
                     }
+                    Fragment::AuthData { .. } => {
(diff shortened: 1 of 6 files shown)


src/send.rs
```diff
@@ -122,6 +122,9 @@ impl<K> SendCommandState<K> {
                         });
                         break true;
                     }
+                    Fragment::AuthData { .. } => {
+                        unimplemented!()
```
duesee, line 126: This is new in `imap-types` and will be implemented during #71

PR #78, opened 2023-11-16
feat: Forward command continuation response requests

### LINKED ISSUE #75: feat: Filter capabilities in command continuation request response
We currenlty don't need this but if we forward command continuation request responses, we should not forget this.

### LINKED ISSUE #77: feat: Forward command continuation request response

duesee:
Closes #75, #77 

I already refactored a bit. We can drop irrelevant commits here but I think the refactoring should be fine.

### DIFF
--- a/proxy/src/util.rs
+++ b/proxy/src/util.rs
@@ -2,7 +2,10 @@ use std::{fs::File, io::BufReader, path::Path};
 
 use imap_codec::imap_types::{
     core::NonEmptyVec,
-    response::{Bye, Capability, Code, Data, Greeting, Status, StatusBody, Tagged},
+    response::{
+        Bye, Capability, Code, CommandContinuationRequest, CommandContinuationRequestBasic, Data,
+        Greeting, Status, StatusBody, Tagged,
+    },
 };
 use rustls::{Certificate, PrivateKey};
 use thiserror::Error;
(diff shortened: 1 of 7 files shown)


proxy/src/util.rs
```diff
@@ -77,6 +80,21 @@ pub fn filter_capabilities_in_status(status: &mut Status) {
     }
 }
 
+// Remove unsupported capabilities in command continuation request response.
+pub fn filter_capabilities_in_continuation(continuation: &mut CommandContinuationRequest) {
+    if let CommandContinuationRequest::Basic(basic) = continuation {
+        if let Some(Code::Capability(capabilities)) = basic.code() {
+            let capabilities = filter_capabilities(capabilities.clone());
```
jakoschiko, line 87: Hmm, all these filter methods have a lot of unnecessary clones. Wouldn't be necessary if we refactor them from `fn foo(bar: &mut Bar)` to `fn foo(bar: Bar) -> Bar`. However, low priority.

src/client.rs
```diff
@@ -153,9 +161,13 @@ impl ClientFlow {
                     }
                 }
                 Response::Data(data) => break Some(ClientFlowEvent::DataReceived { data }),
-                Response::CommandContinuationRequest(_) => {
-                    self.send_command_state.continue_command();
-                    break None;
+                Response::CommandContinuationRequest(continuation) => {
+                    if self.send_command_state.command_in_progress().is_some() {
+                        self.send_command_state.continue_command();
```
jakoschiko, line 166: How about `continue_command` returning a `bool` instead of checking `command_in_progress`?

src/client.rs
```diff
@@ -218,6 +230,9 @@ pub enum ClientFlowEvent {
     StatusReceived {
         status: Status<'static>,
     },
+    ContinuationReceived {
+        continuation: CommandContinuationRequest<'static>,
+    },
```
jakoschiko, line 235: IMO we should start documenting stuff. And this variant is worth to be documented because it's a little bit complicated: "Continuation that was not consumed by `ClientFlow`..."

jakoschiko (commented):
Looks surprisingly easy. Good job.

duesee:
Oh, sorry @jakoschiko! I enabled auto-merge but forgot we disabled "require at least one approval".

jakoschiko:
There is only one solution, force-push on main!

PR #85, opened 2023-11-29
feat: Handle closed stream

### LINKED ISSUE #79: perf(proxy): What does the proxy do to kill time?
I see a high CPU load when running the proxy even when there are no events. (I have some ideas what happens but we should investigate.)

### LINKED ISSUE #84: feat: Emit event if stream was closed or read buffer overflowed
Currently we ignore there result of `read_buf`. For example:

```rust
let crlf_result = match find_crlf(&self.read_buffer[self.seen_bytes..], self.crlf_relaxed) {
    Some(crlf_result) => crlf_result,
    None => {
        // No full line received yet, more data needed.
        stream.0.read_buf(&mut self.read_buffer).await?;
        return Ok(None);
    }
};
```

But according to the [documentation of `read_buf`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_buf]) the result value 0 has a special meaning:

> If n is 0, then it can indicate one of two scenarios:
> 1. This reader has reached its “end of file” and will likely no longer be able to produce bytes. Note that this does not mean that the reader will always no longer be able to produce bytes.
> 1. The buffer specified had a remaining capacity of zero.

I think we need to handle both cases by doing this:

- If the buffer is not full yet and `read_buf` returns 0, then we need to emit `{Client,Server}FlowEvent::StreamClosed`.
- If the buffer is full and it does not contain a complete IMAP message, then we need to emit `{Client,Server}FlowEvent::ReadBufferOverflowed`.

Fixes #79

jakoschiko:
Detect a closed stream when writing to or reading from the stream and emit an error.

I decided to add another error variant instead of event variant. It felt more natural and the implementation looks nice.

I also decided to add more stuff to the `stream` module because it helped to reduce code duplication.

With this PR I consistently get the following proxy log messages after closing the client:

```text
ERROR service{name="GMX"}:client{addr=127.0.0.1:36884}: Connection terminated role="s2p" error=Stream was closed
ERROR service{name="GMX"}:client{addr=127.0.0.1:36894}: Connection terminated role="s2p" error=Stream was closed
ERROR service{name="Zoho Mail"}:client{addr=127.0.0.1:54410}: Connection terminated role="c2p" error=Stream was closed
ERROR service{name="Zoho Mail"}:client{addr=127.0.0.1:54418}: Connection terminated role="c2p" error=Stream was closed
```
Also there is no increased CPU usage after closing the client anymore. I tested it 10 times, so I've proved it beyond doubt.

Closes #84
Closes #79

### DIFF
--- a/proxy/src/proxy.rs
+++ b/proxy/src/proxy.rs
@@ -263,7 +263,7 @@ fn handle_client_event(
             error!(role = "c2p", %error, ?discarded_bytes, "Discard client message");
             return ControlFlow::Continue;
         }
-        Err(ServerFlowError::Io(error)) => {
+        Err(ServerFlowError::Stream(error)) => {
             error!(role = "c2p", %error, "Connection terminated");
             return ControlFlow::Abort;
         }
@@ -304,7 +304,7 @@ fn handle_server_event(
             error!(role = "c2p", %error, ?discarded_bytes, "Discard server message");
             return ControlFlow::Continue;
         }
-        Err(ClientFlowError::Io(error)) => {
+        Err(ClientFlowError::Stream(error)) => {
             error!(role = "s2p", %error, "Connection terminated");
             return ControlFlow::Abort;
         }
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -58,7 +58,7 @@ pub enum StreamError {
     /// closed indefinitely or temporarily depend on the actual stream implementation.
     #[error("Stream was closed")]
     Closed,
-    /// A technical error occurred in the underlying stream.
+    /// An I/O error occurred in the underlying stream.
     #[error(transparent)]
     Io(#[from] tokio::io::Error),
 }
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -54,7 +54,7 @@ impl AnyStream {
     }
 }
 
-/// An error that occurred when reading from or write to a [`Stream`].
+/// Error during reading from or writing to a [`Stream`].
 #[derive(Debug, Error)]
 pub enum StreamError {
     /// The operation failed because the stream is closed.
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -1,6 +1,8 @@
 use std::{fmt::Debug, pin::Pin};
 
-use tokio::io::{AsyncRead, AsyncWrite};
+use bytes::BytesMut;
+use thiserror::Error;
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
 // TODO: Reconsider this. Do we really need Stream + AnyStream? What is the smallest API that we need to expose?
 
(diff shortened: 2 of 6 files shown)


proxy/src/proxy.rs
```diff
@@ -304,7 +304,7 @@ fn handle_server_event(
             error!(role = "c2p", %error, ?discarded_bytes, "Discard server message");
             return ControlFlow::Continue;
         }
-        Err(ClientFlowError::Io(error)) => {
+        Err(ClientFlowError::Stream(error)) => {
```
jakoschiko, line 307: Should we log `ClientFlowError::Stream(StreamError::Closed)` with a lower log level?

src/stream.rs
```diff
@@ -15,4 +17,48 @@ impl AnyStream {
...
+            Err(StreamError::Closed)
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Transfers all bytes from the write buffer into the stream.
+    pub async fn write_all(&mut self, write_buffer: &mut BytesMut) -> Result<(), StreamError> {
+        while !write_buffer.is_empty() {
+            let byte_count = self.0.write_buf(write_buffer).await?;
```
jakoschiko, line 38: Previously we were using [`write_all_buf`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_all_buf). According to its documentation there is no way to detect a closed stream. So I decided to use a `while` + [`write_buf`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_buf).

proxy/src/proxy.rs
duesee: Not sure. IMAP is rather specific about how a connection is finalized dictating a logout sequence, the consumption of all unsolicited responses before a TCP close, etc. So, I'm not sure what constitutes an error, really. Shall we open an issue and triage at some later time?

src/stream.rs
```diff
@@ -15,4 +17,48 @@ impl AnyStream {
...
+/// An error that occurred when reading from or write to a [`Stream`].
+#[derive(Debug, Error)]
+pub enum StreamError {
+    /// The operation failed because the stream is closed.
+    ///
+    /// We detect this by checking if the read or written byte count is 0. Whether the stream is
+    /// closed indefinitely or temporarily depend on the actual stream implementation.
+    #[error("Stream was closed")]
+    Closed,
+    /// A technical error occurred in the underlying stream.
```
duesee, line 61: Better say I/O error?

src/stream.rs
```diff
@@ -15,4 +17,48 @@ impl AnyStream {
     pub fn new<S: Stream + 'static>(stream: S) -> Self {
         Self(Box::pin(stream))
     }
+
+    /// Transfers at least one byte from the stream into the read buffer.
```
duesee, line 21: How about ...

```suggestion
    /// Reads bytes into the buffer returning [`StreamError::Closed`] when no bytes were read, e.g., due to EOF.
```

Shall we also return `Ok(byte_count)`?

duesee (approved):
Thanks! I left only some minor suggestions :-)

>  I tested it 10 times, so I've proved it beyond doubt.

Hehe :-) https://www.youtube.com/watch?v=MsqLgeNX3E4

duesee:
We can close #84 but should open another issue reminding us to do something about the unlimited buffers. Alternatively, we keep #84 open and reference this issue. What do you think?

src/stream.rs
```diff
@@ -15,4 +17,48 @@ impl AnyStream {
...
+    pub async fn read(&mut self, read_buffer: &mut BytesMut) -> Result<(), StreamError> {
+        let byte_count = self.0.read_buf(read_buffer).await?;
+
+        if byte_count == 0 {
+            // The result is 0 if the stream reached "end of file" or the read buffer was already
+            // full before calling `read_buf`. Because we use an unlimited buffer we know that
+            // the first case occurred.
+            Err(StreamError::Closed)
+        } else {
+            Ok(())
```
duesee, line 31: Shall we also return `Ok(byte_count)`?

src/stream.rs
```diff
@@ -15,4 +17,48 @@ impl AnyStream {
...
+                // we know that the first case occurred.
+                return Err(StreamError::Closed);
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// An error that occurred when reading from or write to a [`Stream`].
```
duesee, line 52:
```suggestion
/// Error during reading or writing to a [`Stream`].
```

proxy/src/proxy.rs
jakoschiko: I've created issue #86.

src/stream.rs
jakoschiko: I tried another wording.

src/stream.rs
jakoschiko: We don't need it, but why not.

src/stream.rs
jakoschiko: I insistiere to use "reading from"

duesee (approved):

PR #88, opened 2023-11-30
docs: Explain why we send before we receive

jakoschiko:
We stumbled across this topic some time ago. With this PR I add comments that explain why I originally decided to implement "send before receive" in `ClientFlow` and `ServerFlow`. I duplicated the comment because IMO it's more important to have this comment in the right place than to reduce duplication.


duesee (approved):
Thanks!

PR #89, opened 2023-11-30
feat: Bind handles to `ClientFlow` and `ServerFlow`

### LINKED ISSUE #53: chore: Bind `{Command,Response}Handle` to `{Client,Server}Flow`
Currently there is no binding and `client_flow1` can theoretically accept handles from `client_flow2`. We will have the same issue with the `Scheduler`/`TaskManager`/...

`Tag`s could be a related topic. We could make sure to never generate a `Tag` twice.

jakoschiko:
Use a boring `AtomicU64` and a second `u64` to bind the handles to the `ClientFlow` and `ServerFlow` instances. There is a shared implementation to prevent code duplication, hope you like it.

Closes #53


duesee:
A boooring PR that resolves an issue -- great!

Still, I feel that the removal of `*HandleGenerator` is a step back because it allows to misuse `*Handle::new` by providing the same value twice. (Similarily, `HandleGenerator::new` could be called with unrelated atoms.)

Do you think we can make the generation of tokens resistant to misuse? Similar to `TagGenerator`?

Keep in mind that there is at least one idiot with commit rights in this repository. Thus, I will happily make all mistakes you let me do 🙈

jakoschiko:
I'm confused what misuse you want to prevent. Before my PR `ImapClient` could simply call `ClientFlowCommandHandle(42)`.

jakoschiko:
@duesee: I made it more complicated. What do you think?

duesee (approved):

PR #95, opened 2023-12-21
feat: Implement AUTHENTICATE flow

duesee:
Closes #71 

I already merged all minor changes, typos, reformatting, etc. on main. This branch is a cleaned-up (and rebased) version of `git diff duesee_issues_16_v2`.

Multiple things happen here (see commits):

* TaskMap (could be a separate PR)
* AUTHENTICATE for imap-flow
  * Please look at the fix in particular. I think we accidentally introduced a bug during the "Dummy refactoring".
* AUTHENTICATE for proxy (could be a separate PR)
* AUTHENTICATE for tasks (could be a separate PR)

### DIFF
--- a/examples/server_authenticate.rs
+++ b/examples/server_authenticate.rs
@@ -1,0 +1,84 @@
+use std::error::Error;
+
+use imap_codec::imap_types::response::{CommandContinuationRequest, Greeting, Status};
+use imap_flow::{
+    server::{ServerFlow, ServerFlowEvent, ServerFlowOptions},
+    stream::AnyStream,
+    types::CommandAuthenticate,
+};
+use tokio::net::TcpListener;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let (mut server, _) = {
+        let stream = {
+            let listener = TcpListener::bind("127.0.0.1:12345").await?;
+            let (stream, _) = listener.accept().await?;
+            stream
+        };
+
+        ServerFlow::send_greeting(
+            AnyStream::new(stream),
+            ServerFlowOptions::default(),
+            Greeting::ok(None, "server_authenticate (example)")?,
+        )
+        .await?
+    };
+
+    let mut current_authenticate_tag = None;
+
+    loop {
+        let event = server.progress().await?;
+        println!("{event:?}");
+
+        // We don't implement any real SASL mechanism in this example.
+        let pretend_to_need_more_data = rand::random();
+
+        match event {
+            ServerFlowEvent::CommandAuthenticateReceived {
+                command_authenticate: CommandAuthenticate { tag, .. },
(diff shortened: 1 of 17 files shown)


examples/server_authenticate.rs
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    let mut current_authenticate_tag = None;
+
+    loop {
+        let event = server.progress().await?;
+        println!("{event:?}");
+
+        // We don't implement any real SASL mechanism in this example.
+        let pretend_to_need_more_data = rand::random();
```
jakoschiko, line 35: I don't understand this. If `pretend_to_need_more_data` is true, then `authenticate_finish` is never called. Is this a useful example?

src/server.rs
```diff
@@ -271,9 +394,16 @@ pub enum ServerFlowEvent {
         /// Formerly enqueued [`Response`] that was now sent.
         response: Response<'static>,
     },
-    CommandReceived {
-        command: Command<'static>,
+    /// Command received.
+    CommandReceived { command: Command<'static> },
+    /// Command AUTHENTICATE received.
```
jakoschiko, line 399: This doc should mention that either `authenticate_continue` or `authenticate_finish` should be called afterwards.

src/server.rs
```diff
@@ -271,9 +394,16 @@ pub enum ServerFlowEvent {
...
+    CommandAuthenticateReceived {
+        command_authenticate: CommandAuthenticate,
     },
+    /// Continuation to AUTHENTICATE received.
+    ///
+    /// Note: This can either mean `Continue` or `Cancel` depending on `authenticate_data`.
+    AuthenticateDataReceived { authenticate_data: AuthenticateData },
```
jakoschiko, line 406: This doc should mention that either `authenticate_continue` or `authenticate_finish` should be called afterwards.

src/client.rs
```diff
@@ -267,17 +378,36 @@ pub enum ClientFlowEvent {
         /// (e.g. [`Code::Alert`](imap_codec::imap_types::response::Code::Alert)).
         status: Status<'static>,
     },
+    AuthenticateStarted {
+        handle: ClientFlowCommandHandle,
+    },
+    /// Server is requesting (more) authentication data.
```
jakoschiko, line 384: The doc should mention that `authenticate_continue` should be called afterwards.

examples/server_authenticate.rs
duesee: `pretend_to_need_more_data` is randomly assigned in the loop. What should happen is that you get a variety of different SASL flows randomly. If `pretend_to_need_more_data` is true, we send a continuation and ask for more. Then, we proceed due to the client sending `AuthenticateData` below and eventually finish when `pretend_to_need_more_data` is false,

jakoschiko (approved):

PR #109, opened 2024-01-17
chore: Simplify `NoCodec`

jakoschiko:
Simplify our `NoCodec` hack by removing the trait bound from `ReceiveState`.


duesee (approved):
Remove the code.
Remove the code.
Remove ...

PR #110, opened 2024-01-17
chore: Refactor `send_response`

jakoschiko:
I hope this makes the module `send_response` more readable. If you like it I'll continue with `send_command`.


duesee (approved):
Thanks! The names are definitely more readable. I like where this is going :+1:

PR #114, opened 2024-01-28
test: Introduce crate `flow-test`

jakoschiko:
First attempt to introduce a test setup that allows to write tests conveniently.

My design constraints:
- Test should have as few boilerplate as possible
  - We want to write **many** tests
- `ClientFlow` and `ServerFlow` are tested separately
  - It's possible that both of them could have the same bug that can't be observed if tested simultaneously
  - There should be test setups for both of them
  - Each test setup is mocking the other side with very simple code
- A test should not spawn any additional threads
  - This makes debugging easier
- It should be possible to run multiple tests in parallel
  - This is the default for `cargo test`
  - Ports might be a problem
  - Tokio runtime might be a problem

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ tokio = { version = "1.32.0", features = ["macros", "net", "rt", "sync"] }
 [workspace]
 resolver = "2"
 members = [
+    "flow-test",
     "proxy",
     "tag-generator",
     "tasks",
--- a/flow-test/Cargo.toml
+++ b/flow-test/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "flow-test"
+version = "0.1.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+bstr = { version = "1.9.0", default-features = false }
+bytes = "1.5.0"
+imap-codec = "2.0.0"
+imap-flow = { path = ".." }
+imap-types = "2.0.0"
+tokio = { version = "1.32.0", features = ["macros", "net", "rt", "time"] }
+tracing = "0.1.37"
+tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
--- a/flow-test/README.md
+++ b/flow-test/README.md
@@ -1,0 +1,3 @@
+# flow-test
+
+Test harness for writing lightweight unit tests for `imap-flow`.
--- a/flow-test/src/client_tester.rs
+++ b/flow-test/src/client_tester.rs
@@ -1,0 +1,155 @@
+use std::net::SocketAddr;
+
+use bstr::ByteSlice;
+use imap_flow::{
+    client::{ClientFlow, ClientFlowError, ClientFlowEvent, ClientFlowOptions},
+    stream::AnyStream,
+};
+use imap_types::bounded_static::ToBoundedStatic;
+use tokio::net::TcpStream;
+use tracing::trace;
+
+use crate::codecs::Codecs;
+
+/// A wrapper for `ClientFlow` suitable for testing.
+pub struct ClientTester {
+    codecs: Codecs,
+    client_flow_options: ClientFlowOptions,
+    connection_state: ConnectionState,
+}
+
+impl ClientTester {
+    pub async fn new(
+        codecs: Codecs,
+        client_flow_options: ClientFlowOptions,
+        server_address: SocketAddr,
+    ) -> Self {
+        let stream = TcpStream::connect(server_address).await.unwrap();
+        trace!(?server_address, "Client is connected");
+        Self {
+            codecs,
+            client_flow_options,
+            connection_state: ConnectionState::Connected { stream },
+        }
+    }
+
+    pub async fn receive_greeting(&mut self, expected_bytes: &[u8]) {
+        let expected_greeting = self.codecs.decode_greeting(expected_bytes);
+        match self.connection_state.take() {
+            ConnectionState::Connected { stream } => {
+                let stream = AnyStream::new(stream);
+                let (client, greeting) =
+                    ClientFlow::receive_greeting(stream, self.client_flow_options.clone())
+                        .await
+                        .unwrap();
+                assert_eq!(expected_greeting, greeting);
+                self.connection_state = ConnectionState::Greeted { client };
+            }
+            ConnectionState::Greeted { .. } => {
+                panic!("Client is already greeted");
+            }
+            ConnectionState::Disconnected => {
+                panic!("Client is already disconnected");
+            }
+        }
+    }
+
+    pub async fn send_command(&mut self, bytes: &[u8]) {
+        let enqueued_command = self.codecs.decode_command_normalized(bytes);
+        let client = self.connection_state.greeted();
+        let enqueued_handle = client.enqueue_command(enqueued_command.to_static());
+        let event = client.progress().await.unwrap();
+        match event {
+            ClientFlowEvent::CommandSent { handle, command } => {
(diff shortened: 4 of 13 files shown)


test-setup/src/client_setup.rs
```diff
@@ -0,0 +1,185 @@
...
+    rt: Runtime,
+    server: Server,
+    client: Client,
+}
+
+impl ClientSetup {
+    pub fn new(client_flow_options: ClientFlowOptions) -> Self {
+        let rt = Runtime::new();
+
+        let listener = rt.exec(async { TcpListener::bind("127.0.0.0:0").await.unwrap() });
```
duesee, line 35:
```suggestion
        let listener = rt.exec(async { TcpListener::bind("127.0.0.1:0").await.unwrap() });
```

:-)

test-setup/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+bytes = "1.5.0"
+imap-codec = "2.0.0"
+imap-flow = { path = ".." }
+imap-types = "2.0.0"
+tokio = { version = "1.32.0", features = ["full"] }
```
duesee, line 12: suggestion: We could use a more minimal feature set to spare us a bit compile time.

test-setup/src/client_setup.rs
```diff
@@ -0,0 +1,185 @@
...
+        let bytes = self.command_codec.encode(&expected_command).dump();
+
+        self.rt.exec(self.server.read_expected_bytes(&bytes));
+    }
+}
+
+enum Client {
+    NotConnected { stream: TcpStream },
+    Connected { flow: ClientFlow },
+    Dummy,
```
duesee, line 138: question: Would it make sense to use an `Option<Client>` and remove `Client::Dummy`?

test-setup/src/client_setup.rs
```diff
@@ -0,0 +1,185 @@
...
+
+                self.stream.read_buf(&mut self.read_buffer).await.unwrap();
+            } else {
+                let bytes = &bytes[..expected_count];
+
+                assert_eq!(expected_bytes, bytes);
+
+                self.read_buffer.advance(expected_bytes.len());
+                break;
+            }
```
duesee, line 182: question: Is the first `assert_eq!` necessary? From my understanding, this loop would repeatedly compare prefixes (but with a different length), i.e., 

```
aaaa[..1] != a
aaaa[..2] != aa
aaaa[..3] != aaa
aaaa[..4] != aaaa
```

But I'm tired, maybe I'm overlooking something :-)

test-setup/tests/client.rs
```diff
@@ -0,0 +1,13 @@
...
+#[test]
+fn example() {
+    let mut setup = ClientSetup::new(ClientFlowOptions::default());
+
+    let greeting = setup.server_sends_greeting(b"* OK ...\n");
+    setup.client_receives_greeting(greeting);
+
+    let noop = setup.client_sends_command(b"A1 NOOP\n");
+    setup.server_receives_command(noop);
+}
```
duesee, line 13: Nice!

duesee (approved):

test-setup/src/client_setup.rs
jakoschiko: Yes, maybe, this is WIP and I'm not happy with it either.

test-setup/src/client_setup.rs
jakoschiko: The whole purpose of this function is to assert the incoming bytes. The first `assert_eq!` is useful because it would abort earlier. If we expect banana but get only foo, then the second `assert_eq!` would never be reached.

test-setup/Cargo.toml
jakoschiko: Yes

test-setup/tests/client.rs
jakoschiko: Yes

jakoschiko:
The current design looks like:

```rust
#[test]
fn simple_example() {
    let mut setup = ClientSetup::new(ClientFlowOptions::default());

    let greeting = setup.server_sends_greeting(b"* OK ...\n");
    setup.client_receives_greeting(greeting);

    let noop = setup.client_sends_command(b"A1 NOOP\n");
    setup.server_receives_command(noop);
}
```

All actions (send and receive) are sequential and there is no visible async/await in the test. This is very convenient, but it has a caveat: We assume that there is enough TCP buffer between server and client to buffer the message. If we send a very large message, it will fail:

```rust
#[test]
fn very_large_greeting() {
    let mut setup = ClientSetup::new(ClientFlowOptions::default());

    let mut gibberish = vec![b'x' as u8; 9999999];
    gibberish.push(b'\n');

    setup.server_sends_gibberish(&gibberish); // <- Timeout because send buffer is full
    setup.client_receives_error(); // <- This line is never executed
}
```
On my machine it's possible to send messages with up to ~200KB.

What should we do? I think there are 3 options:

1. Keep the current design because it's sufficient for most tests.
2. Keep the current design, but buffer all actions under the hood. The test then needs to call a method `setup.process()` to execute the client and server actions in parallel.
3. Change the current design drastically. First the test needs to create all relevant `Future`s. Then the test joins and executes them with our `Runtime` (which is currently only an implementation detail of `ClientSetup`). The test would have more control, but would also be more verbose.

What do you think?

jakoschiko:
I tried out finding a better design and came up with this:
```rust
#[test]
fn simple_example() {
    let (rt, mut server, mut client) = TestSetup::default().test_client();

    let greeting = b"* OK ...\r\n";
    rt.run2(server.send(greeting), client.receive_greeting(greeting));

    let noop = b"A1 NOOP\r\n";
    rt.run2(client.send_command(noop), server.receive(noop));

    let status = b"A1 OK ...\r\n";
    rt.run2(server.send(status), client.receive_status(status));
}
```
I really like the new design!
- The required line numbers for a unit test has not changed
- It works with arbitrarily large messages
- It has less magic because `ServerMock` has only two methods now, `send` and `receive`

jakoschiko:
I think the current state is a good proof-of-concept that can be merged and it's ready to be reviewed.

duesee (approved):
Looks simple! I like it, too. I made a few improvements (that you can take, drop, or question) and have a design question:

Design:

* Using strings (instead of objects) makes the tests easier to read.
  * Do we expect to use way more complicated messages? They can be created from `message.dump()` so should still be fine.
* We use a canonical representation for IMAP messages (which technically does not exist.) The types in imap-types are somehow the "canonical" representation, though. So, there is a chance we may get failing test because someone in imap-codec decided to use another "defaut" representation :-) This is mostly casing but there are a few instances of `(item)` (with parantheses) v.ss `item` (w/o paranthesis). I think imap-codec always opts for the shortest option. But it could happen that we need to change due to broken servers or so. Anyway, I generally think it's fine.

jakoschiko:
> Do we expect to use way more complicated messages? They can be created from `message.dump()` so should still be fine.

Yes, we'll use more complicated messages. Also I want to generate random messages via `arbitrary` or `dicetest`. `message.dump()` is a good solution.

> We use a canonical representation for IMAP messages (which technically does not exist.)

The mock is allowed to send non-canonical messages. And that's good because we might want to test that we can parse them (in this case the mock would represent a server/client which is not implemented via `imap-codec`).

> The types in imap-types are somehow the "canonical" representation, though. So, there is a chance we may get failing test because someone in imap-codec decided to use another "default" representation :-)

I think that's okay because it will happen rarely.

PR #117, opened 2024-02-09
refactor: Make option types more backwards compatible

jakoschiko:
We have two option types: `ServerFlowOptions` and `ClientFlowOptions`. We might want to add additional fields to them in the future. It would be nice if we were able to do that without a breaking change. So I propose:

- Let's make them `#[non_exhaustive]`. This allows us to add fields.
- Let's make `ClientFlowOptions` non-`Copy`. This allos us to add non-`Copy` fields.

The `Codec` types in `imap-codec` already do this.

Caveat: To my surprise the `..` notation does not work with `#[non_exhaustive]` (maybe in the future, see https://github.com/rust-lang/rust/issues/70564#issuecomment-1409770088). So this doesn't work:
```rust
let options = ServerFlowOptions {
  crlf_relaxed: false,
  ..ServerFlowOptions::default()
}
```
Instead you need to write this:
```rust
let mut options = ServerFlowOptions::default();
options.crlf_relaxed = false;
```
But I'm willing to accept this.

### DIFF
--- a/proxy/src/proxy.rs
+++ b/proxy/src/proxy.rs
@@ -194,7 +194,7 @@ impl Proxy<ConnectedState> {
     pub async fn start_conversation(self) {
         let (mut proxy_to_server, mut greeting) = {
             // TODO: Read options from config
-            let options = ClientFlowOptions { crlf_relaxed: true };
+            let options = ClientFlowOptions::default();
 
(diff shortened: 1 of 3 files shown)


duesee (approved):
Sounds good! Thanks!

PR #131, opened 2024-02-20
refactor: Optimize line parsing

### LINKED ISSUE #128: Line parsing has O(n²) complexity
It looks like our implementation of `ReceiveState` has O(n²) complexity for line parsing. We can see this by running the test `noop_with_large_lines` with different line sizes.

I ran the test with:

```text
cargo +nightly test --package flow-test --test flow-test-client noop_with_large_lines -- -Z unstable-options --report-time
```

Sample 1
```rust
const LARGE: usize = 10 * 1024 * 1024;
```
```text
test noop_with_large_lines ... ok <1.212s>
```

Sample 2
```rust
const LARGE: usize = 100 * 1024 * 1024;
```
```text
test noop_with_large_lines ... ok <36.650s>
```

Sample 3
```rust
const LARGE: usize = 1000 * 1024 * 1024;
```
```
^C
```

There was already a TODO in `receive.rs` for this problem, but we forgot to create an issue.

```rust
// TODO: If the line is really long and we need multiple attempts to receive it, then this is O(n^2).
//       This could be fixed by setting seen bytes in the None case
let crlf_result = match find_crlf(&self.read_buffer[self.seen_bytes..], self.crlf_relaxed) {
    Some(crlf_result) => crlf_result,
    None => {
        // No full line received yet, more data needed.
        stream.read(&mut self.read_buffer).await?;
        return Ok(None);
    }
};
```

The conversation between server and client looks like this during the test run:

```text
server sends:  * OK ...
client parses: * OK ...
server sends:  .......
client parses: * OK ..........
server sends:  .......
client parses: * OK .................
server sends:  .......
client parses: * OK ........................
server sends:  .......
client parses: * OK ...............................
server sends:  .......
client parses: * OK ......................................
```

Note that this problem exists in both directions. So a client can use large lines for an efficient DOS attack. Luckily the fix should be quite easy, the TODO already suggests it.

jakoschiko:
Reduce the complexity of line parsing from O(n²) to O(n).

Tested with:
```
cargo +nightly test --package flow-test --test flow-test-client noop_with_large_lines -- -Z unstable-options --report-time
```

Sample 1
```rust
const LARGE: usize = 10 * 1024 * 1024;
```
```
test noop_with_large_lines ... ok <0.862s>
```

Sample 2
```rust
const LARGE: usize = 100 * 1024 * 1024;
```
```
test noop_with_large_lines ... ok <8.781s>
```

Sample 3
```rust
const LARGE: usize = 1000 * 1024 * 1024;
```
```
test noop_with_large_lines ... ok <90.486s>
```

Closes #128.

### DIFF
--- a/src/receive.rs
+++ b/src/receive.rs
@@ -23,7 +23,7 @@ impl<C> ReceiveState<C> {
         Self {
             codec,
             crlf_relaxed,
-            next_fragment: NextFragment::default(),
+            next_fragment: NextFragment::start_new_line(),
             seen_bytes: 0,
             read_buffer,
         }
@@ -37,7 +37,7 @@ impl<C> ReceiveState<C> {
     pub fn finish_message(&mut self) {
         self.read_buffer.advance(self.seen_bytes);
         self.seen_bytes = 0;
-        self.next_fragment = NextFragment::default();
+        self.next_fragment = NextFragment::start_new_line();
     }
 
     pub fn discard_message(&mut self) -> Box<[u8]> {
@@ -54,8 +54,8 @@ impl<C> ReceiveState<C> {
     {
         loop {
             match self.next_fragment {
-                NextFragment::Line => {
-                    if let Some(event) = self.progress_line(stream).await? {
+                NextFragment::Line { seen_bytes_in_line } => {
+                    if let Some(event) = self.progress_line(stream, seen_bytes_in_line).await? {
                         return Ok(event);
                     }
                 }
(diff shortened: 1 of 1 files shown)


duesee (approved):
Perfect, thanks!

My understanding of this issue:

Let's assume we have a `max_line_limit` of 1 MiB, i.e., a million bytes.

Attacker sends 1 MiB (w/o `\n`) and then byte, byte, byte, ... This means that sending a single byte makes our implementation check a million bytes. I think about this as "amplification of a million".

Now, we havn't said what the `max_line_limit` should be. I remember reading about a minimum of 8 KiB. So, this would be an amplification of 8000.

I guess what I want to say is: With a limit of 8 KiB, it still doesn't sound too bad. If an attacker wants to make us parse 8000 bytes, they can send 8000 bytes.

Is my understanding correct?

Having said that... Let's *of course* remove any amplification we can find! So, I'm super happy you already fixed this one. The other one could be tough... but more interesting! :-)

PS: I was a bit surprised that parsing 100MiB of IMAP data took 36 (and now 8) seconds... Don't know, I expected at least under a second. Maybe due to `debug`? I guess we will have some surprises in the benchmark milestone.

jakoschiko:
> Is my understanding correct?

Yes.

> Having said that... Let's _of course_ remove any amplification we can find! So, I'm super happy you already fixed this one. The other one could be tough... but more interesting! :-)

Yes.

> PS: I was a bit surprised that parsing 100 MiB of IMAP data took 36 (and now 8) seconds... Don't know, I expected at least under a second. Maybe due to `debug`? Do you mind running with `--release` on your machine for a better intuition? I guess we will have some surprises in the benchmark milestone.

Tested with:
```
cargo +nightly test --release --package flow-test --test flow-test-client noop_with_large_lines -- -Z unstable-options --report-time
```

Sample 1
```rust
const LARGE: usize = 10 * 1024 * 1024;
```
```
test noop_with_large_lines ... ok <0.025s>
```

Sample 2
```rust
const LARGE: usize = 100 * 1024 * 1024;
```
```
noop_with_large_lines ... ok <0.292s>
```

Sample 3
```rust
const LARGE: usize = 1000 * 1024 * 1024;
```
```
test noop_with_large_lines ... ok <2.669s>
```

Issue #118, opened 2024-02-12
Add more `integration-test` tests


jakoschiko:
We need to write tests using the `integration-test` framework introduced in #114.

- [x] `crtl_relaxed` is true, `\n` can be parsed (#133)
- [x] `ctrl_relaxed` is false, `\n` causes malformed (#133)
- [x] literal was accepted (#119)
- [x] literal was rejected (#119)
- [x] literal with non-sync (#230)
- [x] idle accepted (#218)
- [x] idle rejected (#218)
- [x] authenticate accepted (#227)
- [x] authenticate rejected immediately (#227)
- [x] authenticate rejected later (#227)
- [ ] ~~does not panic if functions are called randomly~~ delegated to #232
- [x] server and client send in parallel, see #134 (#231)

For each case we need to consider whether we want to test client, server, or both.

jakoschiko:
[arbtest](https://crates.io/crates/arbtest) looks like a good candidate for tests with random data.

duesee:
Uuuuh, looks nice.

jakoschiko:
Hmm, I'm not happy with arbtest. I wasn't able to generate `Vec<Vec<u8>>` with significant lengths. [See this example](https://www.rustexplorer.com/b#LyoKW2RlcGVuZGVuY2llc10KYXJidGVzdCA9ICIwLjMuMSIKKi8KCnVzZSBhcmJ0ZXN0OjphcmJ0ZXN0OwoKZm4gbWFpbigpIHsKICAgIGFyYnRlc3QofHV8IHsKICAgICAgICBsZXQgeDogVmVjPFZlYzx1OD4+ID0gdS5hcmJpdHJhcnkoKT87CiAgICAgICAgcHJpbnRsbiEoInt4Oj99Iik7CiAgICAgICAgT2soKCkpCiAgICB9KQogICAgLmJ1ZGdldF9tcygxMDAwKQogICAgLnNpemVfbWluKDEwMDAwMCkKICAgIC5zaXplX21heCgxMDAwMDAwMDApOwp9Cg==).

```rust
use arbtest::arbtest;

fn main() {
    arbtest(|u| {
        let x: Vec<Vec<u8>> = u.arbitrary()?;
        println!("{x:?}");
        Ok(())
    })
    .budget_ms(1000)
    .size_min(100000)
    .size_max(100000000);
}
```

produces

```
[]
[[]]
[]
[[156], [215], [19, 242, 169], [52, 80]]
[]
[...]

PR #133, opened 2024-02-20
Change default for crlf_relaxed to false

### LINKED ISSUE #116: Change default for `crlf_relaxed` to false
You mentioned that you want to get rid of the `quirk_crlf_relaxed` feature flag in `imap-codec` at some point. Until then it would be wise to change the default value of `ServerFlowOptions::crlf_relaxed` and `ClientFlowOptions::crlf_relaxed` from true to false.

jakoschiko:
Closes #116.

Adds some tests for  #118.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -32,8 +32,8 @@ pub struct ClientFlowOptions {
 impl Default for ClientFlowOptions {
     fn default() -> Self {
         Self {
-            // Lean towards usability
-            crlf_relaxed: true,
+            // Lean towards conformity
+            crlf_relaxed: false,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/client.rs
```diff
@@ -29,11 +29,12 @@ pub struct ClientFlowOptions {
     pub crlf_relaxed: bool,
 }
 
+#[allow(clippy::derivable_impls)]
 impl Default for ClientFlowOptions {
     fn default() -> Self {
         Self {
-            // Lean towards usability
-            crlf_relaxed: true,
+            // Lean towards conformity
```
duesee, line 36: :+1:

duesee (approved):
Awesome, thanks!

PR #136, opened 2024-02-21
security: Ensure that no secrets are `Debug` printed

### LINKED ISSUE #90: Ensure that no secrets are exposed by `Debug` output
If I understand correctly we want to prevent that the `Debug` output of any type contains any secret.

But what about these cases?
- The read and write buffer of `ClientFlow` and `ServerFlow` may contain the bytes of a secret
- The stream inside of `AnyStream` may contain additional buffers and may write them to its `Debug` output
- Error types like `ClientFlowError::MalformedMessage` may contain the bytes of a secret

duesee:
Closes #90 (I intend to squash all this. Commits only for review.)


src/client.rs
```diff
@@ -383,7 +399,7 @@ pub enum ClientFlowError {
     #[error(transparent)]
     Stream(#[from] StreamError),
     #[error("Expected `\\r\\n`, got `\\n`")]
-    ExpectedCrlfGotLf { discarded_bytes: Box<[u8]> },
+    ExpectedCrlfGotLf { discarded_bytes: Secret<Box<[u8]>> },
```
duesee, line 402: Maybe it's better to implement a debug version by hand here. Would spare us some type conversions ...

src/client.rs
```diff
@@ -48,6 +48,14 @@ pub struct ClientFlow {
...
 }
 
+impl Debug for ClientFlow {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        f.debug_struct("ClientFlow")
+            .field("handle_generator", &self.handle_generator)
+            .finish_non_exhaustive()
+    }
+}
+
```
jakoschiko, line 58: I don't think it's necessary, but you can keep it. But why only for `ClientFlow` and not `ServerFlow`?

src/client.rs
jakoschiko: No, `Secret` is fine for me.

src/stream.rs
```diff
@@ -27,6 +26,7 @@ impl AnyStream {
         let old_len = read_buffer.len();
         let byte_count = self.0.read_buf(read_buffer).await?;
 
+        #[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
duesee: Uff. I'm spoiled by the *awesome* `imap-codec` CI. Will fix (and check why CI didn't catch it.)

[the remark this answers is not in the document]
duesee: https://github.com/duesee/imap-flow/issues/139

PR #137, opened 2024-02-21
refactor!: fix wrong usage of the term `continuation`

duesee:
I'm still not happy about our names but now the semantics are right.


jakoschiko (approved):

jakoschiko:
It looks definitely better than before.

PR #141, opened 2024-02-23
feat(proxy): Log handles (+ logging improvements)

### LINKED ISSUE #138: chore(proxy): Fix remaining `TODO`s in `proxy`
Can we fix the (trivial) TODOs before we move it?

_Originally posted by @jakoschiko in https://github.com/duesee/imap-flow/issues/132#issuecomment-1955303725_

duesee:
Closes #138

### DIFF
--- a/proxy/src/proxy.rs
+++ b/proxy/src/proxy.rs
@@ -97,8 +97,8 @@ impl Proxy<BoundState> {
                     config
                 };
 
-                // TODO: The acceptor should really be part of the proxy initialization.
-                //       However, for testing purposes, it's nice to create it on-the-fly.
+                // TODO(#146): The acceptor should really be part of the proxy initialization.
+                //             However, for testing purposes, it's nice to create it on-the-fly.
                 let acceptor = TlsAcceptor::from(Arc::new(config));
 
                 AnyStream::new(acceptor.accept(client_to_proxy).await?)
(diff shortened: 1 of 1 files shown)


proxy/src/proxy.rs
```diff
@@ -206,11 +207,12 @@ impl Proxy<ConnectedState> {
                 }
             }
         };
-        trace!(greeting=%format!("{:?}", greeting).blue(), role = "s2p", "<--| Received greeting");
+        let msg = format!("{:?}", greeting).blue();
```
jakoschiko, line 210: This forces strict evaluation (i.e. the string is created even if the logger is disabled). Do we want to accept that? It might cause large overhead if the proxy is used in production.

Same for all other cases.

proxy/src/proxy.rs
```diff
@@ -206,11 +207,12 @@ impl Proxy<ConnectedState> {
...
+        trace!(role = "s2p", greeting=%msg, "<--|");
 
         util::filter_capabilities_in_greeting(&mut greeting);
 
-        let (mut client_to_proxy, greeting) = {
+        let (mut client_to_proxy, _) = {
             // TODO: Read options from config
```
jakoschiko, line 216: Can you create issues for the remaining TODOs?

jakoschiko (commented):
Finally, thanks.

proxy/src/proxy.rs
duesee: Good catch! No. It was an oversight and I want to fix this before merging.

PR #159, opened 2024-04-11
Fix rustdoc::broken_intra_doc_links

epilys:
Fix warnings that show up when running `cargo doc [..]` which reports warnings like:

```
  warning: unresolved link to `ClientFlow::set_idle_done`
     --> src/send_command.rs:682:25
      |
  682 |     /// Specifically, [`ClientFlow::set_idle_done`].
      |                         ^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `ClientFlow` in scope
      |
      = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -340,7 +340,7 @@ pub enum ClientFlowEvent {
     },
     /// Server is requesting (more) authentication data.
     ///
-    /// The client MUST call [`ClientFlow::authenticate_continue`] next.
+    /// The client MUST call [`SendCommandState::authenticate_continue`] next.
     ///
     /// Note: The client can also progress the authentication by sending [`AuthenticateData::Cancel`].
     /// However, it's up to the server to abort the authentication flow by sending a tagged status response.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@
 pub mod client;
 mod handle;
 mod receive;
-mod send_command;
+pub mod send_command;
 mod send_response;
 pub mod server;
 pub mod stream;
--- a/src/send_command.rs
+++ b/src/send_command.rs
@@ -614,7 +614,7 @@ enum AuthenticateActivity {
     WaitingForAuthenticateResponse,
     /// Waiting until the client flow user provides the authenticate data.
     ///
-    /// Specifically, [`ClientFlow::set_authenticate_data`].
+    /// Specifically, [`SendCommandState::set_authenticate_data`].
     WaitingForAuthenticateDataSet,
     /// Pushing the authenticate data to the write buffer.
     PushingAuthenticateData { authenticate_data: Vec<u8> },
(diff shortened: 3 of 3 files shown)


duesee:
Hey Manos! Thanks 🥰

I'm a bit confused by `cargo doc`. When I run ...

```sh
cargo doc
```

... I only get ...

warning: unresolved link to `ClientFlow::authenticate_continue`
   --> src/client.rs:343:32
    |
343 |     /// The client MUST call [`ClientFlow::authenticate_continue`] next.
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `ClientFlow` has no field or associated item named `authenticate_continue`
    |
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: `imap-flow` (lib doc) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 6.21s
   Generated /home/duesee/repos/email/imap/imap-flow/target/doc/imap_flow/index.html
```

This one I understand. It should be `ClientFlow::set_authenticate_data`. (Reminder: "Check cargo doc in CI").

But: I cannot reproduce the "ClientFlow::set_idle_done" issue. Which is weird because when I skim through the file, `ClientFlow` is not imported? Any idea what is going on?

By the way, and sorry for the confusing docs, we intend to only have `{Client,Server}Flow...` in the API. If there is a need to export `SendState` or so, this is rather a mistake.

TL;DR

I *think* that fixing `ClientFlow::authenticate_continue` -> `ClientFlow::set_authenticate_data` should be enough. Can you say what errors you get in this case?

File: src/lib.rs [hunk 1]
```diff
@@ -2,7 +2,7 @@
 pub mod client;
 mod handle;
 mod receive;
-mod send_command;
+pub mod send_command;
```
duesee, line 5: This should not be `pub`.

src/send_command.rs
```diff
@@ -614,7 +614,7 @@ enum AuthenticateActivity {
     WaitingForAuthenticateResponse,
     /// Waiting until the client flow user provides the authenticate data.
     ///
-    /// Specifically, [`ClientFlow::set_authenticate_data`].
+    /// Specifically, [`SendCommandState::set_authenticate_data`].
```
duesee, line 617: `ClientFlow` is correct, I'm not sure why I don't get a warning, because `ClientFlow `it's not imported? (See comment.)

src/client.rs
```diff
@@ -340,7 +340,7 @@ pub enum ClientFlowEvent {
     },
     /// Server is requesting (more) authentication data.
     ///
-    /// The client MUST call [`ClientFlow::authenticate_continue`] next.
+    /// The client MUST call [`SendCommandState::authenticate_continue`] next.
```
duesee, line 343:
```suggestion
    /// The client MUST call [`SendCommandState::set_authenticate_data`] next.
```

duesee:
> I think that fixing ClientFlow::authenticate_continue -> ClientFlow::set_authenticate_data should be enough. Can you say what errors you get in this case?

Additionally fixing ...

* `ClientFlow::set_idle_done` -> `crate::client::ClientFlow::set_idle_done` and
* `ClientFlow::set_authenticate_data` -> `crate::client::ClientFlow::set_authenticate_data`

... should do the trick :-)

epilys:
I don't think I understand what's the API supposed to be then, so I'm closing this PR.

PR #162, opened 2024-04-13
ci: Port CI to `just`

### LINKED ISSUE #160: ci: Add `cargo doc --document-private-items` to CI

### LINKED ISSUE #161: ci: Port CI to `just`

duesee:
Closes #161, and closes #160.

### Additional changes

* Updated versions
* Increased MSRV to 1.70 (due to `toml` dep in `proxy`)
* Added minimal_versions test (and workaround for `sharded-slab`)
* Collect coverage from `flow-test`

Should be merged after #159 because the CI should checks doc warnings :-)


justfile
jakoschiko, line 1: You are using `cargo install` with and without `--locked`. Is there any pattern?

jakoschiko (approved):

justfile
duesee: Of course there is... I just copied the install commands fom the docs w/o thinking too much :-)

Do we need to be more mindful here?

duesee: I'll resolve this for now. There are a few improvements we want to make to the just file. But, in order to not block this any longer, let's merge and fix upcoming issues one by one. I'll open a tracking issue.

PR #164, opened 2024-04-13
docs: fix `rustdoc::broken_intra_doc_links`

duesee:


src/send_command.rs
```diff
@@ -614,7 +614,7 @@ enum AuthenticateActivity {
     WaitingForAuthenticateResponse,
     /// Waiting until the client flow user provides the authenticate data.
     ///
-    /// Specifically, [`ClientFlow::set_authenticate_data`].
+    /// Specifically, [`crate::client::ClientFlow::set_authenticate_data`].
```
epilys, line 617:
```suggestion
    /// Specifically, [`ClientFlow::set_authenticate_data`](crate::client::ClientFlow::set_authenticate_data).
```

This will hide the full path from the visible link.

src/send_command.rs
```diff
@@ -679,7 +679,7 @@ enum IdleActivity {
     WaitingForIdleResponse,
     /// Waiting until the client flow user triggers idle done.
     ///
-    /// Specifically, [`ClientFlow::set_idle_done`].
+    /// Specifically, [`crate::client::ClientFlow::set_idle_done`].
```
epilys, line 682: Same

PR #165, opened 2024-05-03
Add pre-built tasks

soywod:
I open the PR so you can already have a look at it, but I plan to leave comments here and there to explain the changes.


Cargo.toml
```diff
@@ -10,16 +10,16 @@ expose_stream = []
 [dependencies]
 bounded-static = "0.5.0"
 bytes = "1.5.0"
-imap-codec = { version = "2.0.0", features = ["quirk_crlf_relaxed", "bounded-static"] }
-imap-types = { version = "2.0.0" }
+imap-codec = { version = "2.0.0", features = ["quirk_crlf_relaxed", "bounded-static", "ext_id", "ext_sort_thread", "ext_binary", "ext_uidplus"] }
```
soywod, line 13: Not sure if `imap-flow` should expose the same features as `imap-{types,codec}` or if it should just enable all of them.

src/client.rs
```diff
@@ -152,8 +177,22 @@ impl ClientFlow {
                 return Ok(event);
             }
 
-            if let Some(event) = self.progress_receive().await? {
-                return Ok(event);
+            if self.is_waiting_for_idle_done_set() {
```
soywod, line 180: This is the uggliest part of the PR. For the `IDLE` command to work with the current flow, I acted the same way the `imap` crate does: every time an `IDLE` event is received (or a timeout), we exit the `IDLE` mode. Lib consumers need to take care of the loop.

tasks/src/lib.rs
```diff
@@ -80,22 +135,114 @@ pub trait Task: 'static {
 }
 
 /// Scheduler managing enqueued tasks and routing incoming responses to active tasks.
+#[derive(Debug)]
 pub struct Scheduler {
     flow: ClientFlow,
     waiting_tasks: TaskMap,
     active_tasks: TaskMap,
     tag_generator: TagGenerator,
+    capabilities: Vec1<Capability<'static>>,
```
soywod, line 144: The inner capabilities. They are updated during the flow (during greetings, before auth, after auth etc). It also allows users to ask for capability availability (see function `ext_<capability>_supported()` functions below.

tasks/src/lib.rs
```diff
@@ -80,22 +135,114 @@ pub trait Task: 'static {
...
+    server_id: Option<Vec<(IString<'static>, NString<'static>)>>,
```
soywod, line 145: Current server id parameters (from the `ID` extension). If found it useful to expose it to users?

tasks/src/lib.rs
```diff
@@ -80,22 +135,114 @@ pub trait Task: 'static {
...
 }
 
 impl Scheduler {
-    /// Create a new scheduler.
-    pub fn new(flow: ClientFlow) -> Self {
-        Self {
+    pub async fn try_new_flow(
```
soywod, line 149: This function prevents users to have wrong or missing capabilities. It takes care of the following logic (as discussed with @duesee): take capabilities from the greeting, if not available → explicitly request them.

tasks/src/lib.rs
```diff
@@ -80,22 +135,114 @@ pub trait Task: 'static {
...
+        scheduler.capabilities = if let Some(Code::Capability(capabilities)) = greeting.code {
+            capabilities
+        } else {
+            scheduler.capability().await?
+        };
+
+        Ok(scheduler)
+    }
+
+    pub async fn authenticate<'a>(
```
soywod, line 175: This function also prevents users to have wrong or missing capabilities. It allows users to send an authenticate task, then get the right capabilities (either from the auth task itself, or from explicit request). It also takes care of sending client ids and storing server's ones.

tasks/src/lib.rs
```diff
@@ -120,6 +267,34 @@ impl Scheduler {
         TaskHandle::new(handle)
     }
 
+    pub async fn exec_task<T>(&mut self, task: T) -> Result<T::Output, SchedulerError>
```
soywod, line 270: Simple wrapper around `enqueue_task` that takes a task, enqueues it then awaits for its resolution.

tasks/src/lib.rs
```diff
@@ -234,16 +419,564 @@ impl Scheduler {
...
+                    };
+
+                    // This `unwrap` can't fail because `active_tasks` contains all in-progress `Commands`.
+                    let (_, _, task) = self.active_tasks.remove_by_handle(handle).unwrap();
+
+                    let output = Some(task.process_tagged(body));
+
+                    return Ok(SchedulerEvent::TaskFinished(TaskToken { handle, output }));
+                }
+                ClientFlowEvent::IdleDoneSent { .. } => {
```
soywod, line 443: Not sure how useful is this event?

tasks/src/lib.rs
```diff
@@ -234,16 +419,564 @@ impl Scheduler {
...
-                ClientFlowEvent::IdleCommandSent { .. } => todo!(),
-                ClientFlowEvent::IdleAccepted { .. } => todo!(),
-                ClientFlowEvent::IdleRejected { .. } => todo!(),
-                ClientFlowEvent::IdleDoneSent { .. } => todo!(),
+                ClientFlowEvent::IdleCommandSent { handle, .. } => {
+                    // This `unwrap` can't fail because `waiting_tasks` contains all unsent `Commands`.
+                    let (handle, tag, task) = self.waiting_tasks.remove_by_handle(handle).unwrap();
+                    self.active_tasks.push_back(handle, tag, task);
+                }
+                ClientFlowEvent::IdleAccepted { .. } => {
```
soywod, line 427: Not sure how useful is this event?

soywod (commented):
Here some comments to give more context. I have the following questions:

- In this PR I missed `STARTTLS` related stuff. I tried to set it up but I have hard time to do so because of `AnyStream` holding a `dyn Stream`. Here a concrete example: I create a `TcpStream`, then wrap it into a `AnyStream`. From there I can create a `ClientFlow` and a `Scheduler`. After the greetings, if `STARTTLS` is supported by the server, I need to take back the initial `TcpStream` in order to upgrade it to a `TlsStream`. Is the SansIO refactor will have an impact on this?

- Regarding SansIO, how is the state of the refactor? Should I wait for it, then re-open a PR?

- Would you like me to split now the PR into small ones? One first with the core, then few others (one per task)?

duesee:
Hey Clement! Thanks for the PR! :-)

I'm still not home but just a few thoughts to (hopefully) unblock you a bit.

I *think* that we have roughly a few work packages here: Tasks that can already be added, tasks that need changes (IDLE), (START)TLS, and other improvements.

To better focus on any part, we could try to split this PR accordingly and first try to get all things in that are unrelated to SansIO. Most tasks, I guess?

We should also treat STARTTLS as a connection "prefix", and only progress the IMAP protocol to the point where both parties start sending TLS data (doing (almost) no parsing).

We don't even need to check capabilities:

* read greeting (no IMAP parsing, greeting has no literal, just read and discard line as we MUST forget everything we learn here anyway)
* send `A starttls` (don't care if supported)
* receive line (don't care if ok or not)
* start with TLS (exactly as an implicit TLS client)

This way, we have an isolated STARTTLS prefix and can proceed exactly as with implicit TLS. This would be secure by design. More in section 8 https://www.usenix.org/conference/usenixsecurity21/presentation/poddebniak.

Jakob is currently working on TLS. I would suggest to wait a few more days before coming back to it.

src/client.rs
```diff
@@ -111,11 +123,24 @@ impl ClientFlow {
...
         Ok((client_flow, greeting))
     }
 
+    pub fn set_some_idle_timeout(&mut self, secs: Option<u64>) {
+        self.idle_timeout = secs;
+    }
+
+    pub fn set_idle_timeout(&mut self, secs: u64) {
+        self.set_some_idle_timeout(Some(secs));
+    }
```
jakoschiko, line 138: Is it necessary to set the timeout after constructing `ClientFlow`? Otherwise I would prefer to move it to `ClientFlowOptions`.

src/send_command.rs
```diff
@@ -19,6 +19,7 @@ use crate::{
     types::CommandAuthenticate,
 };
 
+#[derive(Debug)]
```
jakoschiko, line 22: Adding `Debug` implementations is definitely something that we should do and discuss in a separate PR. In the past there were concerns about not exposing secrets, e.g. see #90.

src/send_command.rs
```diff
@@ -191,7 +202,7 @@ impl SendCommandState {
     pub fn set_authenticate_data(
         &mut self,
         authenticate_data: AuthenticateData<'static>,
-    ) -> Result<ClientFlowCommandHandle, AuthenticateData> {
+    ) -> Result<ClientFlowCommandHandle, AuthenticateData<'static>> {
```
jakoschiko, line 205: Should I make a small PR for fixing this on the main branch?

src/client.rs
```diff
@@ -152,8 +177,22 @@ impl ClientFlow {
                 return Ok(event);
             }
 
-            if let Some(event) = self.progress_receive().await? {
-                return Ok(event);
+            if self.is_waiting_for_idle_done_set() {
+                let timeout =
+                    tokio::time::timeout(self.get_idle_timeout(), self.progress_receive());
+                match timeout.await {
```
jakoschiko, line 183: Please note that the `Future` here can be cancelled. E.g. our proxy implementation cancels the `progress` `Future` all the time because it uses tokio's `select` macro. Then the timer would be started again with the whole timeout interval.

src/client.rs
jakoschiko: Interesting. No strong opinions yet, but some thoughts:

- Not sure whether we should add this to `ClientFlow` or burden the user of `ClientFlow` with implementing this feature. I don't know the pros and cons yet. Or can we implement this as a task?
- Regarding the SansIO refactoring, I would prefer to treat time as an input and somehow pass it into `ClientFlow`. This would also make testing more feasible.
- If we want to implement this as a feature of `ClientFlow`, we should do this in a separate PR and add documentation and `flow-test` test cases.

Cargo.toml
jakoschiko: I have no opinion here. @duesee what do you think?

src/client.rs
```diff
@@ -393,6 +436,8 @@ pub enum ClientFlowEvent {
 pub enum ClientFlowError {
     #[error("Stream was closed")]
     StreamClosed,
+    #[error("Queue receiver was closed")]
+    QueueClosed(#[source] mpsc::error::SendError<()>),
```
jakoschiko, line 440: Where is this error emitted? Can't find it.

Cargo.toml
duesee: No strong opinion either. Let's enable everything for now. Otherwise CI also needs to test some permutations...

soywod:
Awesome, then I will open small PRs until you come back.

> We should also treat STARTTLS as a connection "prefix", and only
> progress the IMAP protocol to the point where both parties start
> sending TLS data (doing (almost) no parsing).

It is more clear now, thank you for the post as well, really
interesting.

> Jakob is currently working on TLS. I would suggest to wait a few more
> days before coming back to it.

I will wait for Jakob's work then.

Thanks for your feedback!

src/client.rs
soywod: I did not dare to change this options, as I was not sure about its scope. So it can definitely move there!

src/send_command.rs
soywod: Sure, it will reduce noise on PRs. Let me know when it's done.

src/client.rs
soywod: It is in the hidden bloc (because diff too big), see https://github.com/duesee/imap-flow/pull/165/files#diff-b9990512dda829fb6d9ef34c240dce3fd5dfd967b089a973a0e54db5784cfe82R725.

src/client.rs
soywod: Do I have to do sth specific here to handle future cancellation?

src/client.rs
soywod: Alright, then I will wait for the SansIO refactoring, and submit a dedicated PR.

src/client.rs
jakoschiko: Ah, okay. Then it would be better to make it a variant of `SchedulerError` instead of `ClientFlowError`.

src/client.rs
jakoschiko: I think there is no general solution for that. In my case It was sufficient to avoid local variables when reaching an `.await` point. I saved the local state somewhere inside of `ClientFlow` or `ServerFlow`. But your case is more complicated because time is involved.

However, with SansIO `ClientFlow` or `ServerFlow` won't use any `Future`s anymore. So let's wait and think about a solution later.

src/client.rs
jakoschiko: Okay! If you want then I can work on this, too.

src/send_command.rs
jakoschiko: Done #166

jakoschiko:
fyi: SansIO is ready for review, see #158

src/client.rs
soywod: I see that the new implem with SansIO reads and writes in parallel, so I'm curious to give it a try and see how it behaves.

soywod:
> fyi: SansIO is ready for review

Awesome, I will try to integrate my core changes from your branch to
see. As soon as it is merged, I will open the first PR (and close this
one).

duesee:
Thanks! :-) Lot's of progress happening, great :-)

PR #172, opened 2024-05-07
Task runner (mid-level API), part 1

soywod:
Split version of #165, part 1.

- Added core changes
- Added the new `Resolver` flow that aims to enqueue task AND get resolution straight. `Resolver::run_task` returns a new `Flow` called `TaskResolver` that can feed `Stream::progress`.
- Made sure `Scheduler` and `Resolver` are `Send`.

### DIFF
--- a/tasks/Cargo.toml
+++ b/tasks/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 imap-flow = { path = ".." }
 imap-types = "2.0.0"
+static_assertions = "1.1.0"
 tag-generator = { path = "../tag-generator" }
 thiserror = "1.0.58"
 tokio = { version = "1.37.0", features = ["net"] }
--- a/tasks/Cargo.toml
+++ b/tasks/Cargo.toml
@@ -4,12 +4,12 @@ version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 
-[dev-dependencies]
-tokio = { version = "1.32.0", features = ["full"] }
-
 [dependencies]
 imap-flow = { path = ".." }
 imap-types = "2.0.0"
 tag-generator = { path = "../tag-generator" }
 thiserror = "1.0.58"
 tokio = { version = "1.37.0", features = ["net"] }
+
+[dev-dependencies]
+tokio = { version = "1.37.0", features = ["full"] }
--- a/tasks/Cargo.toml
+++ b/tasks/Cargo.toml
@@ -7,10 +7,10 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 imap-flow = { path = ".." }
 imap-types = "2.0.0"
-static_assertions = "1.1.0"
 tag-generator = { path = "../tag-generator" }
 thiserror = "1.0.58"
-tokio = { version = "1.37.0", features = ["net"] }
+tracing = "0.1.40"
 
 [dev-dependencies]
+static_assertions = "1.1.0"
 tokio = { version = "1.37.0", features = ["full"] }
--- a/tasks/Cargo.toml
+++ b/tasks/Cargo.toml
@@ -4,9 +4,12 @@ version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 
+[dev-dependencies]
+tokio = { version = "1.32.0", features = ["full"] }
+
 [dependencies]
 imap-flow = { path = ".." }
 imap-types = "2.0.0"
 tag-generator = { path = "../tag-generator" }
 thiserror = "1.0.58"
-tokio = { version = "1.37.0", features = ["full"] }
+tokio = { version = "1.37.0", features = ["net"] }
--- a/tasks/examples/client-tasks.rs
+++ b/tasks/examples/client-tasks.rs
@@ -25,7 +25,7 @@ async fn main() {
     println!("capabilities: {capabilities:#?}");
 
     let handle2 = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));
-    let handle3 = scheduler.enqueue_task(LogoutTask::new());
+    let handle3 = scheduler.enqueue_task(LogoutTask::default());
 
     loop {
         match stream.progress(&mut scheduler).await.unwrap() {
--- a/tasks/examples/client-tasks.rs
+++ b/tasks/examples/client-tasks.rs
@@ -4,7 +4,7 @@ use imap_flow::{
 };
 use imap_types::response::{Response, Status};
 use tasks::{
-    tasks::{authenticate::AuthenticateTask, logout::LogoutTask},
+    tasks::{authenticate::AuthenticateTask, capability::CapabilityTask, logout::LogoutTask},
     Scheduler, SchedulerEvent,
 };
 use tokio::net::TcpStream;
@@ -17,11 +17,10 @@ async fn main() {
     let mut scheduler = Scheduler::new(client);
 
     let capabilities = stream
-        .progress(scheduler.capability())
+        .progress(scheduler.run_task(CapabilityTask::new()))
         .await
         .unwrap()
         .unwrap();
-
     println!("capabilities: {capabilities:#?}");
 
     let handle2 = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));
--- a/tasks/examples/client-tasks.rs
+++ b/tasks/examples/client-tasks.rs
@@ -4,7 +4,7 @@ use imap_flow::{
 };
 use imap_types::response::{Response, Status};
 use tasks::{
-    tasks::{authenticate::AuthenticateTask, capability::CapabilityTask, logout::LogoutTask},
+    tasks::{authenticate::AuthenticateTask, logout::LogoutTask},
     Scheduler, SchedulerEvent,
 };
 use tokio::net::TcpStream;
@@ -16,28 +16,16 @@ async fn main() {
     let client = ClientFlow::new(ClientFlowOptions::default());
     let mut scheduler = Scheduler::new(client);
 
-    let handle1 = scheduler.enqueue_task(CapabilityTask::default());
+    let capabilities = stream
+        .progress(scheduler.capability())
+        .await
+        .unwrap()
+        .unwrap();
 
-    loop {
-        match stream.progress(&mut scheduler).await.unwrap() {
-            SchedulerEvent::TaskFinished(mut token) => {
-                if let Some(capability) = handle1.resolve(&mut token) {
-                    println!("handle1: {capability:?}");
-                    break;
-                }
-            }
-            SchedulerEvent::Unsolicited(unsolicited) => {
-                println!("unsolicited: {unsolicited:?}");
-
-                if let Response::Status(Status::Bye { .. }) = unsolicited {
-                    break;
-                }
-            }
-        }
-    }
+    println!("capabilities: {capabilities:#?}");
 
     let handle2 = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));
-    let handle3 = scheduler.enqueue_task(LogoutTask::default());
+    let handle3 = scheduler.enqueue_task(LogoutTask::new());
 
     loop {
         match stream.progress(&mut scheduler).await.unwrap() {
--- a/tasks/examples/client-tasks.rs
+++ b/tasks/examples/client-tasks.rs
@@ -21,20 +21,21 @@ async fn main() {
         .await
         .unwrap()
         .unwrap();
-    println!("capabilities: {capabilities:#?}");
 
-    let handle2 = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));
-    let handle3 = scheduler.enqueue_task(LogoutTask::default());
+    println!("capabilities: {capabilities:?}");
+
+    let auth_handle = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));
+    let logout_handle = scheduler.enqueue_task(LogoutTask::default());
 
     loop {
         match stream.progress(&mut scheduler).await.unwrap() {
             SchedulerEvent::TaskFinished(mut token) => {
-                if let Some(auth) = handle2.resolve(&mut token) {
-                    println!("handle2: {auth:?}");
+                if let Some(auth) = auth_handle.resolve(&mut token) {
+                    println!("auth: {auth:?}");
                 }
 
-                if let Some(logout) = handle3.resolve(&mut token) {
-                    println!("handle3: {logout:?}");
+                if let Some(logout) = logout_handle.resolve(&mut token) {
+                    println!("logout: {logout:?}");
                     break;
                 }
             }
--- a/tasks/src/lib.rs
+++ b/tasks/src/lib.rs
@@ -17,6 +17,7 @@ use imap_types::{
     core::Tag,
     response::{Bye, CommandContinuationRequest, Data, Response, Status, StatusBody, Tagged},
 };
+use static_assertions::assert_impl_all;
 use tag_generator::TagGenerator;
 use tasks::{
     authenticate::AuthenticateTask, capability::CapabilityTask, logout::LogoutTask, noop::NoOpTask,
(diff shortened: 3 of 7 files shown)


src/client.rs
```diff
@@ -81,12 +82,27 @@ impl ClientFlow {
...
             receive_state,
         }
     }
 
+    /// Changes the actual receive state codec with the
+    /// [`ResponseCodec`]'s one.
+    ///
+    /// The main use case of this function is straight after receiving
+    /// greeting, in order to replace the actual greeting codec by the
+    /// default response codec.
```
jakoschiko, line 97: You don't have to do this. `ClientFlow` already does this by default. See https://github.com/duesee/imap-flow/blob/e24687913ba0761235e6f1e863b17c732ff8b807/src/client.rs#L153-L157

tasks/src/lib.rs
```diff
@@ -26,11 +27,11 @@ use thiserror::Error;
 /// and move out uninteresting responses (returning `Some(...)`).
 ///
 /// If no active task is interested in a given response, we call this response "unsolicited".
-pub trait Task: 'static {
+pub trait Task: Send + 'static {
```
jakoschiko, line 30: Why is `Send` necessary?

jakoschiko:
I don't understand why we need a `Resolver` that wraps the `Scheduler`.

Suggestion: Let's add this method to `Scheduler`:

```rust
impl Scheduler {
    pub fn run_task<'a, T: Task>(&'a mut self, task: T) -> TaskRunner<'a, T> {
        let task_handle = self.enqueue_task(task);
        TaskRunner { scheduler: self, task_handle }
    }
}
```

`TaskRunner` looks like this:

```rust
pub struct TaskRunner<'a, T: Task> {
    scheduler: &'a mut Scheduler,
    task_handle: TaskHandle<T>,
}

impl<'a, T> Flow for TaskRunner<'a, T> {
    type Event = T::Output;
    type Error = SchedulerError;

    fn enqueue_input(&mut self, bytes: &[u8]) {
        self.scheduler.enqueue_input(bytes);
    }

    fn progress(&mut self) -> Result<Self::Event, FlowInterrupt<Self::Error>> {
        todo!()
    }
}
```

Then we can use it like this:

```rust
async fn main() {
    let stream = TcpStream::connect("127.0.0.1:12345").await.unwrap();
    let mut stream = Stream::insecure(stream);

    let client = ClientFlow::new(ClientFlowOptions::default());
    let mut scheduler = Scheduler::new(client);

    let capabilities = stream
        .progress(scheduler.run_task(CapabilityTask::default()))
        .await
        .unwrap();

    println!("capabilities: {capabilities:?}");

    stream
        .progress(scheduler.run_task(AuthenticateTask::plain("alice", "pa²²w0rd")))
        .await
        .unwrap();

    stream
        .progress(scheduler.run_task(LogoutTask::default()))
        .await
        .unwrap();
}
```

If we want to run tasks in parallel, we could even provide something like this:

```rust
let (foo, bar) = stream
    .progress(scheduler.run_task_2(FooTask::default(), BarTask::default()))
    .await
    .unwrap();
```

And we could play a little bit with the API to make it more readable:

```rust
let foo = stream
    .progress(FooTask::default().run(scheduler)) // Swap order to reduce nesting
    .await
    .unwrap();
```

Pros:

- No additional `Resolver` type
- Instead of duplicating the API and adding many functions to `Resolver`, we can reuse the existing `Task` constructors

What do you think?

src/client.rs
soywod: Oh, great! I missed it, thank you.

tasks/src/lib.rs
soywod: I test my changes directly from `email-lib`, and somehow it was complaining about `Send` missing. Since my use case is really basic (I believe it should be the case of most users), I thought it would not hurt to add the `Send` restriction.

soywod:
I initially tried sth similar, but the compiler was not happy at all about the lifetime introduction. Let me try again.

soywod:
Awesome, looks like it work well this time. I did not think about creating a new struct *from the scheduler itself* that returns self, somehow the compiler is happy. Thank you, I prefer this version more!

soywod:
(Regarding the CI, I wait for #175 to be merged to fix it)

tasks/src/lib.rs
jakoschiko: > I test my changes directly from `email-lib`, and somehow it was complaining about `Send` missing.

I think you can write `T: Task + Send` in your lib in places where it's necessary.

> Since my use case is really basic (I believe it would be the case of most users), I thought it would not hurt to add the `Send` restriction since `this trait is automatically implemented when the compiler determines it’s appropriate` [[sources](https://doc.rust-lang.org/std/marker/trait.Send.html)].

Yes, all current implementors of `Task` are `Send`. But adding this bound

- implies that `imap-tasks` somehow relies on this bound, which is not true and
- prevents us from adding implementors in the future that are not `Send` (e.g. because they borrow data instead of owning them).

If there is still a good reason to add this bound, than I'm okay with it. But if you can solve your problem by writing `T: Task + Send` in `email-lib`, then I think we should prefer that. If you point me to your problem in `email-lib` then maybe I can better understand the nature of the problem.

jakoschiko:
I'm on vacation for a couple of days. I'll provide some feedback next week.

tasks/src/lib.rs
soywod: I dug a bit more, and the issue comes from places where `Box<dyn Any>` is used, mostly `TaskAny::process_tagged` and `TaskMap::tasks`. In `email-code`, the backend struct that holds a scheduler (soon a client) have the `Send` restriction by nature, which "imposes" this restriction down. From `email-lib` I have no control over this behaviour. Here the minimum changes I needed to do in order to make the compiler happy: https://github.com/duesee/imap-flow/pull/174/commits/0da136d34d08adc455dbae472eecc6d49b796c69.

I don't see any happy path:

- Without `Send` restriction in `imap-tasks`, users working with `Send` restriction cannot use the Scheduler
- With `Send` restriction in `imap-tasks`, users working without `Send` restriction (`Rc`) cannot use the Scheduler
- Users cannot change this behaviour from their side

Any idea @duesee?

jakoschiko: So your problem is that `Scheduler` is not `Send`. I think that's a good reason to add the bound to `Task`. Then we should also add an [assertion](https://docs.rs/static_assertions/latest/static_assertions/macro.assert_impl_all.html) to ensure that `Scheduler` is `Send` and we don't break it accidentally.

tasks/src/lib.rs
```diff
@@ -342,6 +380,52 @@ pub enum SchedulerError {
...
+    UnexpectedByeResponse(Bye<'static>),
+
+    #[error("unexpected BAD response: {}", .0.text)]
+    UnexpectedBadResponse(StatusBody<'static>),
+
+    #[error("unexpected NO response: {}", .0.text)]
+    UnexpectedNoResponse(StatusBody<'static>),
+
+    #[error("missing required data for command {0}")]
+    MissingData(String),
```
duesee, line 394: I think we need to be a bit careful here to avoid already getting in the weeds of IMAP command semantics.

`UnexpectedTaggedResponse` is a `SchedulerError` because it signals that `imap-tasks` lost track of connection understanding. This could be a severe error during scheduling.

The variants we want to add here are rather task specific, though. Maybe we should create a dedicated enum that holds these common errors? This would also allow us to do refinements (as explained above).

I'm not sure yet about `UnexpectedByeResponse`...

tasks/src/tasks.rs
duesee, line 1: Hehe, we should have started with dedicated files per task :-) Thanks for sorting this out!

What I intend to do: I want to move all tasks into dedicated files first, and then apply your changes. This way it's easier to see what has changed exactly. Unfortunately, we have no diffing tool yet that is smart enough to detect moves :-( Makes it rather hard to see what changed.

tasks/src/tasks/authenticate.rs
```diff
@@ -0,0 +1,90 @@
...
+                },
+            )),
+            StatusKind::No => Err(SchedulerError::UnexpectedNoResponse(status_body)),
+            StatusKind::Bad => Err(SchedulerError::UnexpectedBadResponse(status_body)),
+        }
+    }
+}
+
+impl AuthenticateTask {
+    pub fn new_plain(login: &str, passwd: &str, ir: bool) -> Self {
```
duesee, line 80: I think we can call this just "plain".

duesee (commented):
Hey Clement, some preliminary comments :-)

tasks/src/lib.rs
```diff
@@ -136,6 +139,40 @@ impl Scheduler {
...
+    /// Run the [`LogoutTask`].
+    pub fn logout(&mut self) -> TaskRunner<LogoutTask> {
+        self.run_task(LogoutTask::new())
+    }
+
+    /// Run the [`NoOpTask`].
+    pub fn noop(&mut self) -> TaskRunner<NoOpTask> {
+        self.run_task(NoOpTask::new())
+    }
+
```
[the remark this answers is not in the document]
soywod: It makes sense, then I will make the high-level API use directly tasks instead.

tasks/Cargo.toml
```diff
@@ -7,6 +7,10 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 imap-flow = { path = ".." }
 imap-types = "2.0.0"
+static_assertions = "1.1.0"
 tag-generator = { path = "../tag-generator" }
 thiserror = "1.0.58"
+tokio = { version = "1.37.0", features = ["net"] }
```
jakoschiko, line 13: Do we need tokio as non-dev dependency? I thought `Scheduler` is sans I/O now.

tasks/examples/client-tasks.rs
```diff
@@ -16,25 +16,12 @@ async fn main() {
...
-    }
+    let capabilities = stream
+        .progress(scheduler.run_task(CapabilityTask::new()))
+        .await
+        .unwrap()
+        .unwrap();
+    println!("capabilities: {capabilities:#?}");
 
     let handle2 = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));
     let handle3 = scheduler.enqueue_task(LogoutTask::default());
```
jakoschiko, line 27: `handle1` is missing now. Let's rename the variables to `authenticate_handle` and `logout_handle`.

tasks/src/lib.rs
```diff
@@ -342,6 +355,42 @@ pub enum SchedulerError {
...
+        self.scheduler.enqueue_input(bytes);
+    }
+
+    fn progress(&mut self) -> Result<Self::Event, FlowInterrupt<Self::Error>> {
+        loop {
+            match self.scheduler.progress()? {
+                SchedulerEvent::TaskFinished(mut token) => {
+                    if let Some(output) = self.handle.resolve(&mut token) {
+                        break Ok(output);
+                    }
```
jakoschiko, line 381: In the long run we should find a better idea how to deal with unexpected tokens here. But for now, what do you think about addding a log statement with WARN level? `imap-flow` already has `tracing` as logging dependency.

tasks/src/lib.rs
```diff
@@ -342,6 +355,42 @@ pub enum SchedulerError {
...
+                }
+                SchedulerEvent::Unsolicited(unsolicited) => {
+                    if let Response::Status(Status::Bye(bye)) = unsolicited {
+                        let err = SchedulerError::UnexpectedByeResponse(bye);
+                        break Err(FlowInterrupt::Error(err));
+                    } else {
+                        println!("unsolicited: {unsolicited:?}");
```
jakoschiko, line 388: Let's better change this `println` to a log statement with WARN level.

jakoschiko:
Hello! Thanks for progressing this PR. We were just discussing it and stumbled upon a design issue that is created by having `enqueue_task` and `run_task` on the `Scheduler`. Let's look at this example:

```rust
let handle = scheduler.enqueue_task(AuthenticateTask::plain("alice", "pa²²w0rd", true));

let capabilities = stream
    .progress(scheduler.run_task(CapabilityTask::new()))
    .await // This will silently ignore the result of the enqueued AuthenticateTask
    .unwrap()
    .unwrap();
```

Having both methods on the same type might lead to accidental misusage. Maybe you were right about introducing a dedicated type called `Resolver` (or any other name).

- `Scheduler` is more low-level and provides only `enqueue_task`. Its usage is complicated but rather flexible.
- `Resolver` is more high-level and provides only `run_task`. Its usage is easy but rather restrictive.

However, these are long-term thoughts. If you are happy with the current state then let's keep it for now and improve the design later. I'm sure I'll change my mind about this a couple of times :p

duesee (approved):
Awesome, thanks! As soon as the remaining discussions are fixed we should quickly merge :-) I'm already accepting, thanks for the perseverence!

tasks/Cargo.toml
soywod: No we don't, it's just dead code from refactors. I will remove it, nice catch!

soywod:
> Having both methods on the same type might lead to accidental misusage. Maybe you were right about introducing a dedicated type called `Resolver` (or any other name).

Indeed, I makes sense to separate them to avoid misuage. I initiated a new module `resolver`, let me know what you think about it.

Ready for another review!

tasks/src/resolver.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    fn progress(&mut self) -> Result<Self::Event, FlowInterrupt<Self::Error>> {
+        self.scheduler.progress()
+    }
+}
+
+impl Resolver {
+    /// Create a new resolver.
+    pub fn new(flow: ClientFlow) -> Self {
```
jakoschiko, line 27: I like that the `Scheduler` is just an implementation of `Resolver` and does not appear in its API.

tasks/src/resolver.rs
```diff
@@ -0,0 +1,87 @@
...
+        Self {
+            scheduler: Scheduler::new(flow),
+        }
+    }
+
+    /// Enqueue a [`Task`] for immediate resolution.
+    pub fn run_task<T: Task>(&mut self, task: T) -> TaskResolver<T> {
```
jakoschiko, line 34: I'm not quite happy with the current names `Resolver`, `Resolver::run_task` and `TaskResolver`.

Some alternatives:

- `Resolver`, `Resolver::resolve`, `ResolvingTask`
- `Runner`, `Runner::run`, `RunningTask`

But I'm just nitpicking. I you like we can keep the current names for now.

tasks/src/resolver.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    fn progress(&mut self) -> Result<Self::Event, FlowInterrupt<Self::Error>> {
+        loop {
+            match self.resolver.progress()? {
+                SchedulerEvent::TaskFinished(mut token) => {
+                    if let Some(output) = self.handle.resolve(&mut token) {
+                        break Ok(output);
+                    } else {
+                        warn!("received unexpected task token: {token:?}")
```
jakoschiko, line 64: When using `tracing` it's more common to separate message and data. 

```suggestion
                        warn!(?token, "received unexpected task token")
```

tasks/src/resolver.rs
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                }
+                SchedulerEvent::Unsolicited(unsolicited) => {
+                    if let Response::Status(Status::Bye(bye)) = unsolicited {
+                        let err = SchedulerError::UnexpectedByeResponse(bye);
+                        break Err(FlowInterrupt::Error(err));
+                    } else {
+                        warn!("received unsolicited: {unsolicited:?}");
```
jakoschiko, line 72:
```suggestion
                        warn!(?unsolicited, "received unsolicited");
```

jakoschiko (approved):

jakoschiko:
Thanks again! I added some more comments, but only minor issues, you can choose whether you want to fix them.

If you are ready, we can merge it.

tasks/src/resolver.rs
soywod: I prefer resolve over run, because it gives this information of running a task till the end. So I opted for the first one :ok_hand:

soywod:
> Thanks again! I added some more comments, but only minor issues, you can choose whether you want to fix them.

Fixes pushed, waiting for the CI. If no more comment then it can be merged straight! I will open another PR with all other tasks.

Thank you :slightly_smiling_face:

tasks/src/resolver.rs
soywod: That said, the resolver still uses scheduler events and errors, is it OK? Should type aliases or struct/enum wrappers be defined?

jakoschiko: It's okay. There are different alternatives how we could proceed and I don't want to decide it now. The solution should probably also include the unexpected cases that we log at the moment.

jakoschiko (approved):

PR #173, opened 2024-05-07
Basic Nix support

soywod:
This PR brings Nix (Flakes) support for development shells. Basically, any Nix user that would like to contribute to the project just needs to `nix-shell` (or `nix develop` with flake support) in order to spawn a shell with the Rust environment ready to use.

Later steps could be to add `apps` and `packages` to build or test the lib. Ideally from the CI, to benefit from the reproducible build environment Nix proposes.

### Useful links

- [Nix/Nixpkgs/NixOS website](https://nixos.org/)
- [Nix/Nixpkgs/NixOS GitHub repo](https://github.com/NixOS)
- [Use direnv to automatically spawn Nix shells](https://github.com/nix-community/nix-direnv)
- [Nix Flakes Wiki](https://nixos.wiki/wiki/Flakes)
- [Nice blog about Nix and Nix Flakes](https://zero-to-nix.com/)


.gitignore
```diff
@@ -6,3 +6,4 @@
 
 # direnv (https://direnv.net/)
 .envrc
+.direnv
```
duesee, line 9: Can you say a bit about this file? I tried to find documentation but failed.

duesee:
Grr... bad luck. First time the CI broke like this. Will look into it. https://github.com/rust-lang/rustup/issues/3709

.gitignore
soywod: Looks like the [layout directory](https://github.com/direnv/direnv/blob/6a73cc2d7d13054829501df8756db04e79ff6a29/stdlib.sh#L121) used by direnv. You can see that direnv itself is ignoring this file ([sources](https://github.com/direnv/direnv/blob/master/.gitignore#L2)). People want to add `.direnv` to gitignore, but somehow it always get refused (https://github.com/github/gitignore/pull/4342, https://github.com/github/gitignore/pull/2075).

duesee: Weird... neither the website nor the man pages mention this. Only understood what this does after looking here https://github.com/direnv/direnv/wiki#project-layouts under Python and Ruby.

It seems there is no Rust layout yet. But ".direnv" is used consistently, so it makes sense to anticipate that 👍🏻

Learned something new about direnv :-)

rust-toolchain.toml
duesee, line 1: Question: After [June 13](https://releases.rs/docs/1.79.0/), 1.78.0 won't be the latest stable version. Also, it's not the MSRV either. What is it? And what should it be?

Does it make sense to use 1.66.0 (because it's imap-flow's MSRV)?

duesee: From my understanding, adding the rust-toolchain.toml has quite a few implications that go beyond basic Nix support.

When we narrow these changes to Nix, we might be able to postpone these questions.

Feedback welcome @jakoschiko :-)

duesee (commented):
Thank you very much, Clement!

I'm a bit torn. On one side, I really want Nix users to feel at home. On the other side, I'm not a (seasoned) Nix user myself and cannot really assess your work, and am worried about maintenance. (I have a bit Nix experience but, for example, cannot tell instantly if the flake.lock should be commited, what the compat does, etc.)

Can you say a sentence about each added file? Also, I hope you will help me out when something needs to be reviewed? :-)

Few more questions inline.

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+
+          defaultShell = pkgs.mkShell {
+            buildInputs = with pkgs; [
+              # Nix
+              nixd
+              nixpkgs-fmt
+
+              # Rust
+              rust-toolchain
+              cargo-watch
```
duesee, line 46: Can we be a bit more mindful here? Is it easy to add more tools to a dev shell? If so, we probably should add only tools here that are required to fully check/build the project, i.e., the things that just will install?

flake.nix
```diff
@@ -0,0 +1,59 @@
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-23.11";
+
+    fenix = {
+      url = "github:nix-community/fenix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+
+    flake-compat = {
```
duesee, line 10: Is this required? Can we assume flake support by now? Related: Is this workaround needed in every project?

soywod:
> I'm a bit torn. On one side, I really want Nix users to feel at home. On the other side, I'm not a (seasoned) Nix user myself and cannot really assess your work. (I have a bit Nix experience but, for example, cannot tell instantly if the flake.lock should be commited, what the compat does, etc.)

I totally understand. I experienced the same feeling when someone sent me a PR for supporting Nix in Himalaya for the first time. Somehow it boosted me to learn about, and finally adopt it, but not everyone is willing to do so.

I propose to give you as much information as I can, they you guys decide to accept or not the PR?

> Can you say a sentence about each added file? Also, I hope you will help me out when something needs to be reviewed? :-)

Sure, let me do this right now.

rust-toolchain.toml
soywod: Short answer: we can replace the pinned version with `stable`.

Long answer: Nix needs to know in detail every inputs to be able to properly reproduce environments. It needs to be predictable. If we put `stable` in `rust-toolchain.toml`, the [hash](https://github.com/duesee/imap-flow/pull/173/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R35) will change everytime a new version is out (which will break the CI and forces you to update manually the hash). I was tired of this behaviour, so I ended up pinning a version on Himalaya. I prefer to check myself new releases, check compatibility then manually update the version in `rust-toolchain.toml` with its associated hash. So it's up to you.

Side note on why using `rust-toolchain.toml` with Nix: you maximize changes to have predictable behaviour among users. Both rustup and Nix users use the same source of trust that is the `rust-toolchain.toml` config file. It prevents rustup users to use nightly version while Nix users remaining on stable.

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+
+          rust-toolchain = fenix.packages.${system}.fromToolchainFile {
+            file = ./rust-toolchain.toml;
+            sha256 = "opUgs6ckUQCyDxcB9Wy51pqhd0MPGHUVbwRKKPGiwZU=";
+          };
+
+          defaultShell = pkgs.mkShell {
+            buildInputs = with pkgs; [
+              # Nix
+              nixd
```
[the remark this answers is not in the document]
soywod: It's one of the main stream Nix language server for Nix. It can definitely be removed if you want a minimalist version. It just convenient for Nix (+ direnv) users to enter the project and have a LSP ready to use, including the code formatter.

flake.nix
soywod: Same as `nixd`, it is for convenience. It can be removed. Later on we can add another dev shell with more complete dependencies.

flake.nix
soywod: Nix flakes really improves reproducibility, but it is still (since years…) an experimental features. Therefore it is not available by default in NixOS nor Nixpkgs. The compat is here for non-flakes people to still be able to spawn Nix shells:

- With Flakes support: `nix develop` spawns a shell based on `flake.{nix,lock}`
- Without Flakes support (native Nix): `nix-shell` spawns a shell based on `shell.nix` (or, if missing, `default.nix`).

flake.nix
soywod, line 1: The main file for Nix Flakes users where you can declare dev shells, applications and packages. For now I just declared a default dev shell exposing Rust env (based on rustup's `rust-toolchain.toml` toolchain declaration).

flake.lock
soywod, line 1: The lock file of `flake.nix` (a bit like `Cargo.lock`). Can be updated with `nix flake update`.

Nothing to deal with, but should `Cargo.lock` be committed?

rust-toolchain.toml
soywod, line 1: Rust toolchain configuration. Used by both rustup users and Nix users. Switching to stable channel is more commonly used, yet can lead to [hash](https://github.com/duesee/imap-flow/pull/173/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R39) issue

shell.nix
soywod, line 1: The main file for Nix users. It allows non-Flakes users to still be able to spawn their Nix shell with `nix-shell`.

soywod:
@duesee any idea on this mysterious CI error <https://github.com/duesee/imap-flow/actions/runs/9001307954/job/24727273648?pr=173>?

duesee:
> @duesee any idea on this mysterious CI error https://github.com/duesee/imap-flow/actions/runs/9001307954/job/24727273648?pr=173?

Looks like a bug in cargo-deny? I'm on the CI right now. Will merge the just stuff, then come back to this :-)

duesee (approved):
Thank you very much, Clement! Also for your patience and great explanations. I did a bit of research myself and start feeling more confident.

Very happy to merge this :-)

I also filed #180.

PR #174, opened 2024-05-07
Session (high-level API)

soywod:
> Hm, maybe I didn't understand your use-case correctly. Do you want to process tasks sequentially or in parallel?

Despite the fact that `imap-flow` abstractions are really good (they give a lot of flexibility), I believe most of the casual users (including myself with `email-lib`) do not care about this Sans IO abstraction. They just need a simple API where they can execute simple tasks, in serie. A bit like does the `imap` crate.

So far there is:

- `ClientFlow`, the low-level flow for managing commands
- `Scheduler`, the mid-level flow for scheduling tasks, based on `ClientFlow`
- `Resolver`, also mid-level flow but for resolving tasks one by one, based on `Scheduler` (see #172)

What I try to propose here is the high-level `Session`, composed of a `Stream` and `Resolver`. Users could just connect to a server (TCP), ugrade to TLS and then execute actions like the one proposed here `list_capabilities`.

I would love feedback on the idea.


duesee:
Merged all tasks so far! Now here comes the grand finale? :-)

soywod:
> Merged all tasks so far! Now here comes the grand finale? :-)

Haha exactly. Let me rebase and clean a bit, I let you know when it is
ready for review and discussion.

soywod:
I think it is reviewable (just misses https://github.com/duesee/imap-flow/pull/193), but I don't know how to proceed. The file `client/src/lib.rs` is giant. A big part is opinionated. I mostly "designed" the client according to what `email-lib` needs, but it definitely needs a review. Are you willing to maintain such client inside `imap-flow`?

duesee:
That was fast 🚀

PR #175, opened 2024-05-07
Upgrade `rustls` and `tokio-rustls`

soywod:
I bumped `rustls@0.23.1` and `tokio-rustls@0.26.0`.

I also replaced `webpki-roots` by `rustls-native-certs` (see [why](https://github.com/rustls/rustls-native-certs?tab=readme-ov-file#should-i-use-this-or-webpki-roots)), and put native certs loading in a static [`Lazy`](https://crates.io/crates/once_cell) cell.

---

What is the prefered method to test examples (the server `127.0.0.1:12345`)?

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -45,6 +45,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Setup | Install NASM (Windows)
+        uses: ilammy/setup-nasm@v1
+        if: matrix.os == 'windows-latest'
+
       - name: Setup | Install toolchain
         run: |
           rustup update --no-self-update stable
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -47,8 +47,9 @@ jobs:
 
       - name: Setup | Install toolchain
         run: |
-          rustup toolchain install stable --profile minimal
-          rustup toolchain install nightly --profile minimal
+          rustup update --no-self-update stable
+          rustup update --no-self-update nightly
+          rustup set profile minimal
 
       - name: Setup | Install cargo-fuzz
         run: |
@@ -101,8 +102,9 @@ jobs:
       - name: Setup | Install toolchain
         run: |
           # 1.66 is the Minimum Supported Rust Version (MSRV) for imap-flow.
-          rustup toolchain install 1.66 --profile minimal
-          rustup toolchain install nightly --profile minimal
+          rustup update --no-self-update 1.66
+          rustup update --no-self-update nightly
+          rustup set profile minimal
 
       - name: Setup | Cache dependencies
         uses: Swatinem/rust-cache@v2.5.1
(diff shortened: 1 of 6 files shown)


duesee:
Awesome, thank you! Are you able to fix the remaining CI issues? Skimming over the jobs, we need at least to add "BSD-3-Clause" to deny.toml (and check with cargo deny check)? Happy to have this in. We should also consider using rustls-native-certs in the proxy :-)

duesee:
I think the other CI fail was spurious. I rerun the job, let's see what happens :-) (Happy to have a "local CI" soon :D)

soywod:
> Are you able to fix the remaining CI issues?

Yes, except for the Windows one. I will fix tomorrow morning.

> We should also consider using rustls-native-certs in the proxy :-)

Is it not what I tried to achieve [there](https://github.com/duesee/imap-flow/pull/175/files#diff-b09a3ca57f63ac6f386216d76a6511f8000c5b817b589a2515f8d522a0b5bd0c)?

duesee:
> Is it not what I tried to achieve [there](https://github.com/duesee/imap-flow/pull/175/files#diff-b09a3ca57f63ac6f386216d76a6511f8000c5b817b589a2515f8d522a0b5bd0c)?

Urgh... sorry. I think I confused something. This is even better :-)

soywod:
> What is the prefered method to test examples (the server `127.0.0.1:12345`)?

And what about this? How do you usually spawn your server to test locally?

jakoschiko:
> > What is the prefered method to test examples (the server `127.0.0.1:12345`)?
> 
> And what about this? How do you usually spawn your server to test locally?

The examples can be tested by using netcat. Though it's very time-consuming and I wouldn't recommend it.

Usually we test `imap-flow` by using the proxy. We set up different dummy mail accounts and connect to them via the proxy. The setup is an one-time effort and then it's really easy to do some real-world testing.

We also spent much time in writing integration tests, just run `cargo test` in the `flow-test` folder. Unfortunately the code coverage is not perfect yet and this won't test TLS specific code.

duesee:
What Jakob said. Also: The idea was that you can run client and server in pairs. Maybe we should put that im CI, too. Would require some well-defined condition such as a logout. I'll file an issue.

soywod:
I try to fix the CI (had an idea) but it needs approval every time I push, is there a way to "auto-approve" (at least for this PR)?

duesee:
> I try to fix the CI (had an idea based on [this comment](https://github.com/rust-lang/rustup/issues/3409#issuecomment-1647699079)) but it needs approval every time I push, is there a way to "auto-approve" (at least for this PR)?

Fixed (hopefully). We don't strictly require any approval for PRs anymore :-)

soywod:
Awesome, seems to work! The thing I learnt is that `rustup` is installed by default on GitHub machines. By running `rustup install` it tries to remove the existing installation, and failed on Windows due to permission issues (wrong path?). Better just `rustup update` to avoid this.

Now there is a new error on Windows, I try to fix!

soywod:
Yay, all green! The second error was linked to `tokio-rustls`. It uses `aws-lc` as a crypto lib, which needs [NASM](https://aws.github.io/aws-lc-rs/requirements/windows.html#nasm) on Windows.

Ready to merge.

duesee:
Awesome, thanks! One last thing: Could you cleanup the commit history, i.e., squash all CI changes, and squash all TLS changes? Then this is ready to go :-) Already approving.

soywod:
> Could you cleanup the commit history

Done!

duesee:
Peeeerfect! 🎉 Thanks!

PR #176, opened 2024-05-07
Refactor names of send and receive related types

jakoschiko:
After the refactoring of #158 some type and file names are a little bit confusing. I tried to improve the situation (hopefully for the last time):

Here the new structure
- `client.rs` contains `ClientFlow`
- `client_receive.rs` contains `ClientReceiveState`
- `client_send.rs` contains `ClientSendState`
- `server.rs` contains `ServerFlow`
- `server_receive.rs` contains `ServerReceiveState`
- `server_send.rs` contains `ServerSendState`
- `receive.rs` contains `ReceiveState` which is used by both `ClientReceiveState` and `ServerReceiveState`

Also I tried to replace the specific terms `command` and `response` with the less specific term `message` in many places because the more specific term was sometimes not correct anymore (e.g. `ServerSendState` handles both `Response` and `Greeting`, so the term `message` is more appropriate than `response`).

I split this refactoring in many small commits, so it should be easy to understand the changes.


duesee (approved):
Thanks!

PR #183, opened 2024-05-15
Add `StoreTask`

soywod:


tasks/src/tasks/store.rs
```diff
@@ -0,0 +1,110 @@
...
+            kind: self.kind,
+            response: StoreResponse::Answer,
+            flags: self.flags.clone(),
+            uid: self.uid,
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Fetch { items, seq } = data {
+            self.output.insert(seq, items);
```
duesee, line 39: Can we emit a warning when we get duplicate `seq`s?

tasks/src/tasks/store.rs
```diff
@@ -0,0 +1,110 @@
...
+            None
+        } else {
+            Some(data)
+        }
+    }
+
+    fn process_tagged(self, status_body: StatusBody<'static>) -> Self::Output {
+        match status_body.kind {
+            StatusKind::Ok => Ok(self.output),
```
duesee, line 48: I think we overlooked that before: Here, in the `Status`, there is potentially also a `Code` we want to return. But: This is the case for all `Task`s. Sometimes, the `Code` also makes the `Text` relevant, e.g., `Code::Alert`.

I think we should maybe just ignore this for now. If it is somehow relevant, we can add `StoreTaskV2`.

tasks/src/tasks/store.rs
```diff
@@ -0,0 +1,110 @@
...
+            StatusKind::No => Err(TaskError::UnexpectedNoResponse(status_body)),
```
duesee, line 49: Just to be clear: The `Code` can potentially also be used here...

tasks/src/tasks/store.rs
```diff
@@ -0,0 +1,110 @@
...
+    flag::{Flag, StoreResponse, StoreType},
+    response::{Data, StatusBody, StatusKind},
+    sequence::SequenceSet,
+};
+
+use super::TaskError;
+use crate::Task;
+
+#[derive(Clone, Debug)]
+pub struct StoreTask {
```
duesee, line 16: > The STORE command alters data associated with a message in the
      mailbox.  Normally, STORE will return the updated value of the
      data with an untagged FETCH response.  A suffix of ".SILENT" in
      the data item name prevents the untagged FETCH, and the server
      SHOULD assume that the client has determined the updated value
      itself or does not care about the updated value.

Maybe as doc: `Alter message data.`?

tasks/src/tasks/store.rs
```diff
@@ -0,0 +1,110 @@
...
+        self.set_uid(uid);
+        self
+    }
+
+    pub fn silent(self) -> SilentStoreTask {
+        SilentStoreTask::new(self)
+    }
+}
+
+/// Same as [`StoreTask`], except that it does not return any output.
```
duesee, line 80: Maybe as doc

```
Alter message data instructing the server to not send the updated values.

Note: Same as [`StoreTask`], except that it does not return any output.
```

duesee (changes requested):
Thanks! A few nits and one request and then happy to get this :-)

Thanks for splitting the tasks, btw. Really great to review now.

tasks/src/tasks/store.rs
soywod: What about making `process_tagged` return a `TaskOutput<'static, Self::Output>` instead:

```rust
pub struct TaskOutput<'a, T> {
    pub output: T,
    pub code: Option<Code<'a>>,
    pub text: Text<'a>,
}
```

Since it impacts all actual PRs, let's wait to merge all of them, and then open a new PR with those changes?

tasks/src/tasks/store.rs
soywod: But in this case it is already exposed? I mean, if a client wants to display information from `Code` or `Text`, it has to pattern match the `Result` and check by itself the `status_body`?

tasks/src/tasks/store.rs
duesee: Yeah, let's think later :+1:

tasks/src/tasks/store.rs
duesee: Ah, yes. This was just a reminder for us that errors also potentially have codes. All good here :-)

duesee (approved):
Thanks! 💪🏻

PR #184, opened 2024-05-15
Add `SortTask` and `ThreadTask`

soywod:


tasks/src/tasks/sort.rs
```diff
@@ -0,0 +1,84 @@
...
+        CommandBody::Sort {
+            sort_criteria: self.sort_criteria.clone(),
+            charset: self.charset.clone(),
+            search_criteria: self.search_criteria.clone(),
+            uid: self.uid,
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Sort(ids) = data {
```
duesee, line 36: Grrrrrrr... Do we care about duplicate `Data::Sort`? I think we shouldn't but maybe emit a log again...? What do you think? (Maybe also a topic for later when we merged everything so far.)

tasks/src/tasks/sort.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn process_tagged(self, status_body: StatusBody<'static>) -> Self::Output {
+        match status_body.kind {
+            StatusKind::Ok => Ok(self.output),
+            StatusKind::No => Err(TaskError::UnexpectedNoResponse(status_body)),
+            StatusKind::Bad => Err(TaskError::UnexpectedBadResponse(status_body)),
+        }
+    }
+}
+
+impl SortTask {
```
duesee, line 53: Thought: Maybe we want to unify all tasks later to use the builder pattern. Simple changes we can discuss later. Let's first try to get all tasks in :-)

duesee (approved):
Thanks!

tasks/src/tasks/sort.rs
soywod: Sure, I like the idea :+1:

tasks/src/tasks/sort.rs
soywod: Emitting a log is good enough I would say. I pushed sth, let me know.

PR #185, opened 2024-05-15
Add `FetchTask` and `SearchTask`

soywod:


tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,108 @@
...
+        CommandBody::Fetch {
+            sequence_set: self.sequence_set.clone(),
+            macro_or_item_names: self.macro_or_item_names.clone(),
+            uid: self.uid,
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Fetch { items, seq } = data {
+            self.output.insert(seq, items);
```
duesee, line 35: Can we emit a warning on duplicates here?

tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,108 @@
...
+        self
+    }
+
+    pub fn first(self) -> FetchFirstTask {
+        FetchFirstTask(self)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct FetchFirstTask(FetchTask);
```
duesee, line 76: Can we add a doc?

tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,108 @@
...
+    core::Vec1,
+    fetch::{MacroOrMessageDataItemNames, MessageDataItem},
+    response::{Data, StatusBody, StatusKind},
+    sequence::SequenceSet,
+};
+
+use super::TaskError;
+use crate::Task;
+
+#[derive(Clone, Debug)]
```
duesee, line 14: Can we add a short doc? :-)

tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,116 @@
...
+            sequence_set: self.sequence_set.clone(),
+            macro_or_item_names: self.macro_or_item_names.clone(),
+            uid: self.uid,
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Fetch { items, seq } = data {
+            if let Some(items) = self.output.insert(seq, items) {
+                warn!(?items, "received duplicate items for message {seq}");
```
duesee, line 39: Should we make `seq` also a structured value?

tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.0.command_body()
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        match data {
+            Data::Fetch { items, seq } if self.0.output.is_empty() => {
+                self.0.output.insert(seq, items);
+                None
+            }
+            data => Some(data),
```
duesee, line 105: Ah, I see. If I understand correctly, the remaining `Fetch`s will become `unsolicited` data now? I think this could be confusing...

What I also find not ideal is that this tasks seems to fetch at most one message but really it fetches more.

Maybe it's better to do the filtering at a later stage for now? This would make obvious what happens (and reduce duplicated code for now)? Alternatively, we should maybe think about avoiding the overfetches? Sadly, I don't know how to do this in a single task.

What do you think?

tasks/src/tasks/fetch.rs
soywod: Sth like this?

```diff
- warn!(?items, "received duplicate items for message {seq}");
+ warn!(?items, seq, "received duplicate items");
```

I hesitatetd, somehow I found the non structured version more readable, but I don't have solid argument for. Up to you, what do you think is better?

tasks/src/tasks/fetch.rs
soywod: I pushed an alternative, what do you think about? If it does not fit, I agree to filter before and after, around the `FetchTask`, it also makes sense. I just found convenient the fact to propose a `FetchFirstTask` (somehow I had this need at higher level).

tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn command_body(&self) -> CommandBody<'static> {
+        CommandBody::Fetch {
+            sequence_set: self.sequence_set.clone(),
+            macro_or_item_names: self.macro_or_item_names.clone(),
+            uid: self.uid,
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Fetch { items, seq } = data {
```
soywod, line 37: Here I have a proposition. It may be a bit too opinated, yet I found it relevant to share.

If you use `uid` for this task, the output still returns a hash map of sequence. It could be nice to return either UID or seq depending on `uid: bool`, but it implies to modify the item names provided by the user (pushing the `Uid` item name). Here how it looks like:

```rust
// First expand items, then push Uid if needed
let mut fetch_items = match fetch_items {
    MacroOrMessageDataItemNames::Macro(m) => m.expand().into_static(),
    MacroOrMessageDataItemNames::MessageDataItemNames(items) => items,
};

if uid {
    fetch_items.push(MessageDataItemName::Uid);
}

// In process_data, we can now extract the Uid from the items
fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
    if let Data::Fetch { items, seq } = data {
        if self.uid {
            let uid = items.as_ref().iter().find_map(|item| {
                if let MessageDataItem::Uid(uid) = item {
                    Some(*uid)
                } else {
                    None
                }
            });

	    // We could even unwrap since we know that MessageDataItemName::Uid
	    // have been pushed to the message item names
	    
            if let Some(uid) = uid {
                self.output.insert(uid, items);
            } else {
                warn!("cannot get UID of message {seq}, skipping fetch data");
            }
        } else {
            self.output.insert(seq, items);
        }

        None
    } else {
        Some(data)
    }
}
```

tasks/src/tasks/fetch.rs
duesee: I see why you hesitated. Still, I think I would prefer the structured version.

tasks/src/tasks/fetch.rs
duesee: I like the idea but it has some downsides :-(

The data we get for sure is the `seq`. Everything else might not be present -- even if we ask for it (This is also why we should not use `unwrap`: A broken (or even malicious) server must not be able to crash us.) Thus, what's left is to not include the UID in the hash map (and maybe notify the user) -- exactly what you propose.

The downside is that now we alter the session. Instead of saying "UID missing", we say "fetch not sent". Not sure how important that is, but it feels "surprising". Also, we remove the `seq` which is, I feel, probably better explicitly removed at the user site.

Maybe these downsides can be resolved with a different design. But I feel that the fact that UID may not be present prevents a clean API :-(

tasks/src/tasks/fetch.rs
duesee: I'm a bit tired, so I hope I understand right. The `FetchFirstTask` now has a single seq/uid, right? Is it worth to duplicate the task code for this case? Because: In the end, it's similar to calling `FetchTask` with `seq.iter().first()` instead `seq`? Do I understand right?

What could be a useful task is something like "Fetch the first message from 1:*" (what you had before) BUT with some smarter commands, e.g., first try to find out which message is the first (w/o fetching the others) and then only fetching the first one. But this is currently not possible to do in a single `Task` because it's bound to a single `CommandBody`.

soywod: > Is it the same as calling `FetchTask` with `seq.iter().first()` for a specific `seq`?

Technically yes, but I don't like the fact that running `FetchTask` with one unique specific sequence number or uid (not a range) outputs a hash map. By running `FetchFirstTask`, you can only define 1 seq/uid, therefore you get 1 fetch. In this particular case, I value more the public API over code duplication.

Maybe the naming induces us in error. `FetchFirstTask` implies that you take *the first one from a list*. What about:

- `FetchTask` becomes one of `FetchManyTask`, `RangeFetchTask`
- `FetchFirstTask` becomes one of `FetchTask`, `FetchSingleTask`, `FetchOneTask`

Fetching the first task of a range can be done at the higher level client, as you said.

> But this is currently not possible to do in a single `Task` because `Task` is bound to a single `CommandBody`.

I guess it is sth to do in the higher level client as well?

Sth like could be interesting also is a `FetchAllTask`, which is just a `FetchManyTask` with `..` sequence range.

tasks/src/tasks/fetch.rs
soywod: Makes total sense. What is bothering me the most is the lack of consistency among tasks using either seq or uid. Let's take the `SEARCH`, `SORT`, and `THREAD` tasks: the `uid: bool` parameter affects *input* (defines if sequence set is made of seq or uid) and *output* (returns a vector of seq or uid). In the case of `FETCH` (and `STORE` as well), `uid: bool` only affects *input*: the *output* uses seq no matter what. But after all, it is the way the protocol is defined, so let stick to it.

> Everything else might not be present -- even if we explicitly ask for it.

Sure, but you have the exact same issue if you request manually the uid. A user who `FETCH` using uid implicitly set should be as surprised as the one who `FETCH` using uid explicitly set. So I ask you this question: what is the best way we could handle this case of missing data explicitly (or implitictly) set? Let's say only one `FETCH` does not have uid, should we send all other fetches and notify the missing one, or fail? By experience, I would go for the first option, at least in does not interrupt the user.

Another idea that just came to by mind is to return a struct (or enum) that contains both correct and potentially incorrect output. It is a bit more verbose version of the log warn notification, but at least user has total control of the situation. A good package is [`These`](https://github.com/FintanH/these), the "inclusive or" version of `Result`.

```rust
type Output = Result<These<Vec<MessageDataItem>, Vec<MessageDataItem>>, TaskError>
```

What a mess :smiling_face_with_tear:

tasks/src/tasks/fetch.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn command_body(&self) -> CommandBody<'static> {
+        CommandBody::Fetch {
+            sequence_set: SequenceSet::from(SeqOrUid::from(self.id)),
+            macro_or_item_names: self.macro_or_item_names.clone(),
+            uid: self.uid,
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Fetch { items, .. } = data {
```
duesee, line 119: Sorry for this, but here we should warn on duplicates again to be consistent.

tasks/src/tasks/fetch.rs
duesee: Yeah, it's a mess. Let's discuss this at a later time. `FetchFirstTask` is not too repetitve. Let's get this in amd see how it goes.

duesee (approved):
I think we are at a point where we want to first get all the tasks in and then talk about the current state more globally. I'll file some issues.

So, happy to merge :-) Can you resolve the conflict?

duesee:
Hm... GitHub seems to got in some weird state... Can you try pushing an empty commit or so? Otherwise I'll merge on main manually.

PR #186, opened 2024-05-15
Add mailbox tasks

soywod:


tasks/src/tasks/select.rs
```diff
@@ -0,0 +1,194 @@
...
+    mailbox::Mailbox,
+    response::{Code, Data, StatusBody, StatusKind},
+};
+use tracing::{error, warn};
+
+use super::TaskError;
+use crate::Task;
+
+#[derive(Clone, Debug, Default)]
+pub struct SelectData {
```
soywod, line 15: Should this struct move to `imap-types`?

tasks/src/tasks/expunge.rs
```diff
@@ -0,0 +1,43 @@
...
+    command::CommandBody,
+    response::{Data, StatusBody, StatusKind},
+};
+
+use super::TaskError;
+use crate::Task;
+
+#[derive(Clone, Debug, Default)]
+pub struct ExpungeTask {
+    count: usize,
```
duesee, line 11: I'm not sure I understand the idea of the count. We expect a list of expunged messages, right?

duesee: Ah, maybe adding docs will clear things up :-) I also keep forgetting...

tasks/src/tasks/select.rs
duesee: Not sure... maybe. Hard to tell where the boundaries are.

tasks/src/tasks/select.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+impl Task for SelectTask {
+    type Output = Result<SelectData, TaskError>;
+
+    fn command_body(&self) -> CommandBody<'static> {
+        let mailbox = self.mailbox.clone();
+
+        if self.read_only {
+            CommandBody::Examine { mailbox }
```
duesee, line 76: 👍🏻

tasks/src/tasks/select.rs
```diff
@@ -0,0 +1,194 @@
...
+    ) -> Option<StatusBody<'static>> {
+        let StatusBody { kind, code, text } = status_body;
+
+        match code {
+            Some(Code::Unseen(seq)) => match &kind {
+                StatusKind::Ok => {
+                    self.output.unseen = Some(seq);
+                    None
+                }
+                StatusKind::No => {
```
duesee, line 112: Huch 😕 Didn't think about these cases... I... don't know what to do here. Anything in the RFC? Will try to find out...

duesee (commented):
Thanks, again! Looks really good. For SELECT, I'll need to consult the RFC. Not sure what to do with non-OK status that have a correct code. I feel like `warn`ing and dropping (consuming in task) could make sense but don't know for sure.

tasks/src/tasks/expunge.rs
soywod: I initially wanted to return a `Vec<NonEmptyU32>`, but:

> For example, if the last 5 messages in a 9-message mailbox are expunged, a "lower to higher" server will send five untagged EXPUNGE responses for message sequence number 5, whereas a "higher to lower server" will send successive untagged EXPUNGE responses for message sequence numbers 9, 8, 7, 6, and 5.

Because server behaviour is not well defined, this task could not output the same. It could just output a vector with 5 times the same sequence number… so I ended up just counting the number of expunged messages.

Let me add a tiny doc about it.

duesee: Returning `Vec` allows a consumer to easily do `vec.len()`. But returning `vec.len()` only "removes" potentially useful information. I think we should be careful with this because then it's only a matter of time someone will need to redefine the task (or add another one).

Luckily, it's easy to define custom `Task`s. So, maybe we should try to be a bit more generic for the build-in ones and do more opinionated tasks at higher level?

I think we should avoid dropping information that is already there.

tasks/src/tasks/select.rs
duesee: Okay, I read a bit in the RFC and feel that the most straightforward thing would be to match only exactly what the RFC tells us to match.

For SELECT it says ...

> REQUIRED **OK** untagged responses:  UNSEEN, PERMANENTFLAGS, UIDNEXT, UIDVALIDITY

Thus, I feel like just not processing **NO** and **BAD** and returning as `Some`. These will then show up as unsolicited.

tasks/src/tasks/select.rs
duesee: Thoughts: Ideally, `imap-types` should have only types that are definitely valid by spec. When we add `SelectData`, we would need to do it using `Option` for all fields, which, strictly speaking, is not valid. So, one option would be to call it `SelectData{Unvalidated,Raw}`, etc. But, I think the answer could also be: Keep it in an upper layer.

soywod: > I think the answer could also be: Keep it in an upper layer

But then what would be the return type of this task? And where would be its definition? 

> So, one option would be to call it `SelectData{Unvalidated,Raw}`

It could also be a giant tuple, but I find it uggly. Let's go for `SelectDataUnvalidated`?

tasks/src/tasks/expunge.rs
soywod: I added a better comment to be more explicit.

tasks/src/tasks/select.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    // required OK untagged responses
+    pub unseen: Option<NonZeroU32>,
+    pub permanent_flags: Option<Vec<FlagPerm<'static>>>,
+    pub uid_next: Option<NonZeroU32>,
+    pub uid_validity: Option<NonZeroU32>,
+}
+
+impl SelectDataUnvalidated {
+    pub fn validate(self) -> Result<Self, TaskError> {
```
duesee, line 29: This ...

```rust
impl SelectDataUnvalidated {
    fn validate(self) -> SelectData {}
}
```

... could be a pattern we expose publicly. Right now, `validate` is more like an internal helper (that could take a reference instead of moving.)

I'll add a bullet point in #191. These `Data` structures are worth discussing. Not blocking, let's make progress! :-)

duesee (approved):
Awesome ✨ Thank you very much again!

PR #188, opened 2024-05-15
Add `StartTlsTask`

soywod:
Also adjusted the `LogoutTask` to match style of other tasks.

### DIFF
--- a/tasks/src/tasks/logout.rs
+++ b/tasks/src/tasks/logout.rs
@@ -3,17 +3,12 @@ use imap_types::{
     response::{Bye, StatusBody, StatusKind},
 };
 
-use crate::{tasks::TaskError, Task};
+use super::TaskError;
+use crate::Task;
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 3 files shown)


tasks/src/tasks/logout.rs
duesee, line 1: Hehe, I think we mostly use the `Type` directly followed by `impl` style. So, the `Task`s are different, I think.

Feel free to keep as is or change. Maybe I can do a reordering at some later time to unify the whole code base.

duesee (approved):
Thanks!

tasks/src/tasks/logout.rs
soywod: My bad, [I saw](https://github.com/duesee/imap-flow/blob/d16cda3226a4d687648afe96fa9302b158b47e03/tasks/src/tasks.rs) before implementing tasks that the `impl` was below the `impl Task`, so I naively applied the same logic everywhere.

Let's keep it like this then reorder later in one PR. Sorry for the noise.

PR #189, opened 2024-05-15
Add `AppendTask`, `AppendUidTask` and `CheckTask`

soywod:
Also adjusted `NoOpTask` and `CapabilityTask` to match style of other tasks.

### DIFF
--- a/tasks/src/tasks/append.rs
+++ b/tasks/src/tasks/append.rs
@@ -1,0 +1,179 @@
+use imap_types::{
+    command::CommandBody,
+    datetime::DateTime,
+    extensions::binary::LiteralOrLiteral8,
+    flag::Flag,
+    mailbox::Mailbox,
+    response::{Data, StatusBody, StatusKind},
+};
+use tracing::warn;
+
+use super::TaskError;
+use crate::Task;
+
+#[derive(Clone, Debug)]
+pub struct AppendTask {
+    mailbox: Mailbox<'static>,
+    flags: Vec<Flag<'static>>,
+    date: Option<DateTime>,
+    message: LiteralOrLiteral8<'static>,
+    output: Option<u32>,
+}
+
+impl Task for AppendTask {
+    type Output = Result<Option<u32>, TaskError>;
+
+    fn command_body(&self) -> CommandBody<'static> {
+        CommandBody::Append {
+            mailbox: self.mailbox.clone(),
+            flags: self.flags.clone(),
+            date: self.date.clone(),
+            message: self.message.clone(),
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        // In case the mailbox is already selected, we should receive
+        // an `EXISTS` response.
+        if let Data::Exists(seq) = data {
+            if self.output.is_some() {
+                warn!("received duplicate APPEND EXISTS data");
+            }
+            self.output = Some(seq);
+            None
+        } else {
+            Some(data)
+        }
+    }
+
+    fn process_tagged(self, status_body: StatusBody<'static>) -> Self::Output {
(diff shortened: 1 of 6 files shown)


tasks/src/tasks/append.rs
```diff
@@ -0,0 +1,175 @@
...
+        CommandBody::Append {
+            mailbox: self.mailbox.clone(),
+            flags: self.flags.clone(),
+            date: self.date.clone(),
+            message: self.message.clone(),
+        }
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        // In case the mailbox is already selected, we should receive
```
duesee, line 35: 👍🏻

duesee, line 35: Good catch! :+1:

tasks/src/tasks/append.rs
```diff
@@ -0,0 +1,175 @@
...
+        // an `EXISTS` response.
+        if let Data::Exists(seq) = data {
+            self.output = Some(seq);
```
duesee, line 38: Hm... duplicates again. Maybe we should think about a more generic solution. For now, log should suffice?

tasks/src/tasks/append.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+/// Special [`NoOpTask`](super::noop::NoOpTask) that captures `EXISTS`
+/// responses.
+///
+/// This task should be used whenever [`AppendTask`] does not return
+/// the number of messages in the mailbox the appended message
+/// resides.
+#[derive(Clone, Debug, Default)]
```
duesee, line 99: I'm not sure about this. Need to think a bit...

duesee (commented):
This looks very good so far, thanks! I'm a bit unsure about the `NoOp` variants. Will think about it.

tasks/src/tasks/append.rs
soywod: Yep, I'm not so fan of them either. I literally followed the RFC:

> If the mailbox is currently selected, the normal new message actions SHOULD occur.  Specifically, the server SHOULD notify the client immediately via an untagged EXISTS response.  If the server does not do so, the client MAY issue a NOOP command (or failing that, a CHECK command) after one or more APPEND commands.

The aim behind is to have a similar API with `APPENDUID` (to use as a fallback, in case the extension is not available).

tasks/src/tasks/append.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl Task for PostAppendNoOpTask {
+    type Output = Result<Option<u32>, TaskError>;
+
+    fn command_body(&self) -> CommandBody<'static> {
+        CommandBody::Noop
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Exists(seq) = data {
```
duesee, line 116: Duplicates again. (Just a note, we should go through all tasks im a second step.)

tasks/src/tasks/append.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl Task for PostAppendCheckTask {
+    type Output = Result<Option<u32>, TaskError>;
+
+    fn command_body(&self) -> CommandBody<'static> {
+        CommandBody::Check
+    }
+
+    fn process_data(&mut self, data: Data<'static>) -> Option<Data<'static>> {
+        if let Data::Exists(seq) = data {
```
duesee, line 158: Duplicates. (Only a note.)

tasks/src/tasks/append.rs
duesee: Yeah... Let's get this in :-)

PR #193, opened 2024-05-18
Add support for `XOAUTH2` and `OAUTHBEARER`

soywod:


soywod:
Ready for review!

tasks/src/tasks/authenticate.rs
```diff
@@ -67,27 +92,47 @@ impl Task for AuthenticateTask {
...
-        _: CommandContinuationRequest<'static>,
+        continuation: CommandContinuationRequest<'static>,
     ) -> Result<AuthenticateData<'static>, CommandContinuationRequest<'static>> {
-        if self.ir {
-            Ok(AuthenticateData::Cancel)
-        } else if let Some(line) = self.line.take() {
-            Ok(AuthenticateData::r#continue(line))
+        let data = if self.ir {
+            if let AuthMechanism::XOAuth2 = self.mechanism {
+                // The current continuation request indicates an error,
```
duesee, line 99: Could you explain this a bit more? I feel like I don't understand what is happening here.

From my understanding, the code is describing roughly this trace:

```imap
C: A1 AUTHENTICATE XOAUTH <data>
S: + ... // This must not happen because XOAUTH2 <data> was already send (allowed due to SASL-IR)
// error!("cannot authenticate...")
C: <empty> // line 112
```

Is this what you had in mind? Have you seen this in the real world?

If so, I think the correct approach is to send `AuthenticateData::Cancel`, i.e., 

```imap
C: A1 AUTHENTICATE XOAUTH <data>
S: + ... // This must not happen because XOAUTH2 <data> was already send (allowed due to SASL-IR)
C: * // Here, a single "*" send by the client signals "cancel authentication". It's... IMAP...
```

duesee: What worries me a bit about sending empty data is that there is a chace we end up in an infinite loop.

```
C: A1 AUTHENTICATE XOAUTH <data>
S: + ... // This must not happen because XOAUTH2 <data> was already send (allowed due to SASL-IR)
// error!("cannot authenticate...")
C: <data> // empty data
S: + ... // This must not happen because XOAUTH2 <data> was already send (allowed due to SASL-IR)
// error!("cannot authenticate...")
C: <data> // empty data
S: + ... // This must not happen because XOAUTH2 <data> was already send (allowed due to SASL-IR)
// error!("cannot authenticate...")
C: <data> // empty data
...
```

Sending `AuthenticateData::Cancel` would avoid the loop.

tasks/src/tasks/authenticate.rs
```diff
@@ -67,27 +92,47 @@ impl Task for AuthenticateTask {
...
+                // so we need to send an empty response in order to receive
+                // the final authentication error later on.
+                match continuation {
+                    CommandContinuationRequest::Basic(basic) => {
+                        let text = basic.text();
+                        error!("cannot authenticate using XOAUTH2 mechanism: {text}");
```
duesee, line 105: We could deduplicate line 105 and 109 by putting the log after the match. I'm also not sure if we want to emit an `error!` or a `warn!`ing. An `error!` seems to ask for a an action. But we can't do anything here.

tasks/src/tasks/authenticate.rs
```diff
@@ -67,27 +92,47 @@ impl Task for AuthenticateTask {
...
+                    }
+                    CommandContinuationRequest::Base64(data) => {
+                        let text = String::from_utf8_lossy(data.as_ref());
```
duesee, line 108: We could use `imap-types::utils::escape_byte_string` here to not loose information about the string. It should escape bytes in a reversible way.

duesee (changes requested):
Hello! Thank you very much for the fix and the implementation! I feel like wanting to isolate the mechanisms a bit more by introducing wrapping types so that it's dead easy to see what mechanism is processed. I can do this in a second commit.

Have you tested this code against gmail or so by chance? I feel like I might be missing some information about XOAUTH2.

tasks/src/tasks/authenticate.rs
duesee: Ah, I see why you did it this way :+1:

duesee (approved):
Hey again! I just looked more closely at XOAUTH2 here https://developers.google.com/gmail/imap/xoauth2-protocol

I now understand what happens. I'll add a few more comments but then happy to merge! Thank you very much!

tasks/src/tasks/authenticate.rs
```diff
@@ -67,27 +92,47 @@ impl Task for AuthenticateTask {
...
     fn process_continuation_request_authenticate(
         &mut self,
-        _: CommandContinuationRequest<'static>,
+        continuation: CommandContinuationRequest<'static>,
     ) -> Result<AuthenticateData<'static>, CommandContinuationRequest<'static>> {
-        if self.ir {
-            Ok(AuthenticateData::Cancel)
-        } else if let Some(line) = self.line.take() {
-            Ok(AuthenticateData::r#continue(line))
+        let data = if self.ir {
```
duesee, line 97: Here, we only ask for an error when SASL-IR was uses. But from my reading of the https://developers.google.com/gmail/imap/xoauth2-protocol, we should also handle the case if SASL-IR was not used in a similar way.

duesee:
Update: While rewriting a bit, I discovered that we are missing a few cases. This is why the review takes a bit longer. Have you integrated this already in imap-client? If so, we could proceed from there.

soywod:
> Update: While rewriting a bit, I discovered that we are missing a few cases. This is why the review takes a bit longer. Have you integrated this already in imap-client? If so, we could proceed from there.

I propose to continue and merge this PR first, then let me know your decision about where should be the `tasks` workspace (either inlined inside `imap-client` as a temporary solution, stay here, move in another dedicated repo etc).

As I said, I just initiated the inline solution with `imap-client` to wire up everything in `email-lib`. If you decide sth else I will adjust the code and rewrite git history :slightly_smiling_face:

soywod:
Since you agreed to move tasks to `imap-client`, I close this PR. The code has already been merged [here](https://github.com/soywod/imap-client/blob/main/src/tasks/tasks/authenticate.rs), feel free to open issues or pull requests to adjust.

PR #196, opened 2024-05-27
Implement `Stream` to `TcpStream` conversion

soywod:
Mostly needed for STARTTLS, see usage at https://github.com/duesee/imap-flow/pull/174/files#diff-3f7e738903cc8204a4e3b1c8ab4db5812898e6bb5ad8b6c7b36df42e29481a8fR197

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -172,6 +172,16 @@ impl Stream {
     }
 }
 
+/// Take the [`TcpStream`] out of a [`Stream`].
+///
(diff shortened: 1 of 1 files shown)


soywod:
This is the only missing PR needed by `imap-client`, all other changes can be done directly on `imap-client`.

duesee:
Talked with Jakob and we are not sure we want to expose this. On the other hand, I don't want to block you from getting things done.

Proposal: Can you put this behind the expose_stream feature? Then we merge and evaluate again at a later time.

PR #202, opened 2024-05-30
refactor: Introduce `MessageParser`

### LINKED ISSUE #129: Literal parsing has O(n²) complexity
If `ReceiveState` parses a message with n literals, then the complexity is O(n²). There is already a TODO in `receive.rs`:

```rust
// Try to parse the whole message from the start (including the new line).
// TODO: If the message is really long and we need multiple attempts to receive it, then this is O(n^2)
//       IMO this can be only fixed by using a generator-like decoder
match self.codec.decode(&self.read_buffer[..self.seen_bytes]) {
    Ok((remaining, message)) => {
        assert!(remaining.is_empty());
        Ok(Some(ReceiveEvent::DecodingSuccess(message.into_static())))
    }
    Err(error) => Ok(Some(ReceiveEvent::DecodingFailure(error.into_static()))),
}
```

Unfortunately I think there is no easy fix for that. The parser implemented in `imap-codec` would need to be stateful and remember the partially parsed message in order to solve the issue.

### LINKED ISSUE #198: feat: Support secure discard of invalid messages
Related https://github.com/duesee/imap-codec/issues/432#issuecomment-1962427538

jakoschiko:
In the long term we might want to add a utility to `imap-codec` for better message/line/literal parsing. But before that, let's do some experiments in this repository.

The module `receive` already contains some primitive line/literal parsing. Let's extract and extend this functionality.

Closes #129
Closes #198

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ pub mod client;
 mod client_receive;
 mod client_send;
 mod handle;
+mod message;
 mod receive;
 pub mod server;
 mod server_receive;
--- a/src/message.rs
+++ b/src/message.rs
@@ -1,0 +1,795 @@
+#![allow(unused)]
+
+use std::ops::RangeInclusive;
+
+use imap_codec::decode::Decoder;
+use imap_types::core::LiteralMode;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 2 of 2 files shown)


duesee:
As discussed: Closing in favor of #213.

PR #208, opened 2024-06-01
fix!: Check CCR creation early to avoid panic (or late error handling)

### LINKED ISSUE #170: question: Error during seemingly correct construction of `CommandContinuationRequest`
When writing a test, I used this very common text for accepting literals:

```rust
setup.server_flow_options.literal_accept_text = Text::unvalidated("moah");
```

Unfortunately this caused a panic inside of `imap-flow`:

```
thread 'command_with_literals_larger_than_max_command_size' panicked at src/server.rs:299:38:
called `Result::unwrap()` on an `Err` value: Ambiguity
```

This is the mentioned `unwrap`:

```rust
// Inform the client that the literal was accepted.

// Unwrap: This should never fail because the text is not Base64.
let cont = CommandContinuationRequest::basic(
    None,
    self.options.literal_accept_text.clone(),
)
.unwrap();
self.send_response_state.enqueue_response(
    None,
    Response::CommandContinuationRequest(cont),
);
```

duesee:
First attempt. We could make it work with `&'static str` and `String`... Closes #170.

### DIFF
--- a/flow-test/tests/flow-test-server.rs
+++ b/flow-test/tests/flow-test-server.rs
@@ -1,7 +1,6 @@
 use std::time::Duration;
 
 use flow_test::test_setup::TestSetup;
-use imap_types::core::Text;
 
 #[test]
 fn noop() {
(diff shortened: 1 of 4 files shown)


jakoschiko (approved):
Looks very nice!

PR #213, opened 2024-06-09
refactor: Introduce `Fragmentizer`

### LINKED ISSUE #129: Literal parsing has O(n²) complexity

### LINKED ISSUE #198: feat: Support secure discard of invalid messages

jakoschiko:
In the long term we might want to add a utility to `imap-codec` for better message/line/literal parsing. But before that, let's do some experiments in this repository.

The module `receive` already contains some primitive line/literal parsing. Let's extract and extend this functionality.

Closes #129
Closes #198


Cargo.toml
```diff
@@ -15,11 +15,11 @@ exclude = [
 [features]
 default = ["stream"]
 expose_stream = []
-stream = ["dep:rustls", "dep:tokio", "dep:tokio-rustls"]
+stream = ["dep:bytes", "dep:rustls", "dep:tokio", "dep:tokio-rustls"]
```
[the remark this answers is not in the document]
jakoschiko: `Stream` [is still using it](https://github.com/duesee/imap-next/blob/c0dd3e1c513bb5a7c789e98cfce11a5b7ca5523c/src/stream.rs#L24-L25) for interacting with `TcpStream`.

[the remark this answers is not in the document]
duesee: Ah, my bad.

src/lib.rs
```diff
@@ -1,12 +1,12 @@
 #![forbid(unsafe_code)]
 
 pub mod client;
-mod client_receive;
 mod client_send;
+#[doc(hidden)] // TODO: move to imap-codec and make it public
```
duesee, line 5: We could create an issue and reference it here.

src/fragmentizer.rs
```diff
@@ -0,0 +1,1411 @@
...
+#[derive(Clone, Debug)]
+enum Parser {
+    Line(LineParser),
+    Literal(LiteralParser),
+}
+
+/// Stateful parser for the next line fragment.
+#[derive(Clone, Debug)]
+struct LineParser {
+    // Where we started parsing the line.
```
duesee, line 325:
```suggestion
    /// Where we started parsing the line.
```

src/fragmentizer.rs
```diff
@@ -0,0 +1,1411 @@
...
+    start: usize,
+    // Until where we parsed the line.
```
duesee, line 327:
```suggestion
    /// Until where we parsed the line.
```

src/fragmentizer.rs
```diff
@@ -0,0 +1,1411 @@
...
+    end: usize,
+    // Accumulated state based on the parsed bytes.
```
duesee, line 329:
```suggestion
    /// Accumulated state based on the parsed bytes.
```

src/fragmentizer.rs
```diff
@@ -0,0 +1,1411 @@
...
+    },
+    Cr {
+        announcement: Option<LiteralAnnouncement>,
+    },
+}
+
+/// Stateful parser for the next literal fragment.
+#[derive(Clone, Debug)]
+struct LiteralParser {
+    // Where we started parsing the literal.
```
duesee, line 497:
```suggestion
    /// Where we started parsing the literal.
```

src/fragmentizer.rs
```diff
@@ -0,0 +1,1411 @@
...
+    start: usize,
+    // Until where we parsed the literal.
```
duesee, line 499:
```suggestion
    /// Until where we parsed the literal.
```

src/fragmentizer.rs
```diff
@@ -0,0 +1,1411 @@
...
+    end: usize,
+    // Remaining bytes we need to parse.
```
duesee, line 501:
```suggestion
    /// Remaining bytes we need to parse.
```

duesee (approved):
Thanks! I am very happy with the changes! Spend a bit of time reviewing the tests, and also implemented some fuzz testing (on another branch). I also added a server example that fits here well, and tested if it works as it should.

I propose to merge this and to make a second PR to add fuzz testing and improve the tests. Fuzz testing can easily be moved from imap-next to imap-codec. So, we don't need to worry too much when to bring it in place.

Great work!

src/lib.rs
jakoschiko: I've created #238

PR #214, opened 2024-06-12
fix: Change `Display` text of `Event::CommandTooLong`

jakoschiko:
The current text might imply that the command was somehow automatically rejected, similar to `Error::LiteralTooLong`. But that's not the case. The user is responsible for sending a BAD.


jakoschiko:
I've created #215 for improving the situation.

duesee (approved):

PR #227, opened 2024-06-29
test: Add tests for AUTHENTICATE

jakoschiko:


jakoschiko:
https://datatracker.ietf.org/doc/html/rfc4959

PR #230, opened 2024-06-30
test: Add tests for non-sync literals

jakoschiko:


jakoschiko:
https://datatracker.ietf.org/doc/html/rfc7888

PR #241, opened 2024-07-10
feature(proxy): Log fragments

jakoschiko:


jakoschiko:
Closes #251

duesee:
As discussed, this should now be reopened against [`imap-proxy`](https://github.com/duesee/imap-proxy)

jakoschiko:
See https://github.com/duesee/imap-proxy/pull/13

PR #247, opened 2024-07-12
fix: Handle closed TLS session properly

### LINKED ISSUE #242: perf: `Stream` with TLS runs into infinite loop after connection was closed
The most inner loop in `imap_next::stream::decrypt` is the culprit. It can be tested by using the proxy and then closing the client.

Relates to #79

jakoschiko:
Closes #242

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -276,15 +276,20 @@ fn decrypt(
 
         loop {
             let mut plain_bytes_chunk = [0; 128];
+            // We need to handle different cases according to:
+            // https://docs.rs/rustls/latest/rustls/struct.Reader.html#method.read
             match tls.reader().read(&mut plain_bytes_chunk) {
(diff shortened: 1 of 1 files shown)


src/stream.rs
```diff
@@ -276,15 +276,20 @@ fn decrypt(
...
+                // There are no more bytes to read
+                Err(err) if err.kind() == ErrorKind::WouldBlock => break,
+                // The TLS session was closed uncleanly
+                Err(err) if err.kind() == ErrorKind::UnexpectedEof => {
+                    return Err(DecryptEncryptError::Closed)
                 }
+                // We got an unexpected error
                 Err(err) => return Err(DecryptEncryptError::Io(err)),
+                // The TLS session was closed cleanly
+                Ok(0) => return Err(DecryptEncryptError::Closed),
```
duesee, line 291: Aha! Classic.

duesee (approved):

PR #254, opened 2024-07-13
feature!: Add `max_response_size` for `Client`

### LINKED ISSUE #253: feature: Max response size for `Client`
In #153 we implemented max command size for `Server`. Similarly we could add max response size for `Client`.

Relates to #157

jakoschiko:
Closes #253


src/client.rs
```diff
@@ -34,6 +38,8 @@ impl Default for Options {
         Self {
             // Lean towards conformity
             crlf_relaxed: false,
+            // A value larger than the default `server::Options::max_command_size`.
```
duesee, line 41: I'm not sure I understand this comment. Could you explain? Is the assumption just that responses are generally bigger? 100 MiB should okay, though.

duesee (approved):

src/client.rs
jakoschiko: I just documented how I picked the number. I chose a value greater than the other value. Because I thought the client is usually less constrained than the server.

jakoschiko: I rephrased the comment.

PR #270, opened 2024-09-04
Improve rustls cargo features

soywod:


duesee:
Thanks! So the idea is to make `aws-lc-rs` vs `ring` configurable, right? Did you open this PR on purpose against @jakoschiko feature branch? Or should it have been filed against main?

jakoschiko:
I don't have much experience with feature flags. When does it make sense to forward feature flags like in this PR? I understand why it makes sense for crates that are tightly coupled and do re-exporting, e.g. `imap-types` and `imap-codec`. But why in this case?

If I understand feature flags correctly, you can simply add this to your `Cargo.toml`:

```toml
imap-next = "0.2.0"
tokio-rustls = { version = "*", features = ["ring"] }
```

And then the feature `ring` is activated for `tokio-rustls` because feature flags are additive. This is faster and more flexible than doing PRs in upstream crates. Am I missing something?

Using the re-exported `rustls` is good idea, though.

soywod:
> Did you open this PR on purpose against @jakoschiko's feature branch?

Yes, so it's based on the latest version of `imap-{codec,types}`.

> And then the feature `ring` is activated for `tokio-rustls` because feature flags are additive. This is faster and more flexible than doing PRs in upstream crates. Am I missing something?

Because they are additive, we end up with both `ring` and `aws-lc-rs`. I think it is a common practice, look at [`tokio-rustls`](https://github.com/rustls/tokio-rustls/blob/6f7373dad47df0b071c8f0f1a03376bed08331e7/Cargo.toml#L20-L28). The use case is the following: I'm a final user, and I already use `rustls` + `ring` in my code base. I don't want to build `imap-next` with `rustls` + `aws-lc-rs`. I understood it this way, I may be wrong as well. For sure I had the need, because I tried to fix Nix derivation that was complaining about `aws-lc-rs` even though I was not using in the codebase. Till this PR.

jakoschiko:
Or maybe just remove the default features?

```toml
[features]
default = ["stream"]
# ...

[dependencies]
tokio-rustls = { version = "0.26.0", optional = true, default-features = false }
# ...
```
Then you can use it like this:

```toml
imap-next = { version = "0.2.0" }
tokio-rustls = { version = "*", default-features = false, features = ["logging", "tls12", "ring"] }
```

I can't name any reason why `imap-next` should activate features of `tokio-rustls`.

soywod:
> Or maybe just remove the default features?

I think this should work, let me try.

soywod:
Looks like it works as expected, ready to merge!

PR #274, opened 2024-09-08
refactor!: Reimplement stream without split

jakoschiko:


soywod:
Any news about this refactor? I may need sth similar for [`time-lib`](https://github.com/pimalaya/core/tree/master/time), where a `TcpStream` is used for both reading and writing, and compatible with `tokio` or `async-std`.

jakoschiko:
If it's blocking you, then I'll continue to work on this.

How do think compatibility with `async-std` can be achieved? Currently this PR proposed to depend on `tokio::io::AsyncRead` and `tokio::io::AsyncWrite`. So I think it's not compatible with `async-std`.

soywod:
> If it's blocking you, then I'll continue to work on this.

No don't worry, it's not blocking me. Yet it could be a great addition for few libs of mine, including `imap-client`.

> How do think compatibility with `async-std` can be achieved? Currently this PR proposed to depend on `tokio::io::AsyncRead` and `tokio::io::AsyncWrite`. So I think it's not compatible with `async-std`.

In [`time-lib`](https://github.com/pimalaya/core/blob/master/time/src/tcp.rs) I used `futures::Async{Read,Write}`. I just needed to write some compatibility layer between `tokio` and `async-std`.

I liked your Sans IO approach, I don't know if it could be applicable here, for stream. By default only the stream is exposed, then every feature could expose an IO implementation like `tokio-native-tls`, or `async-std-rustls` etc.

soywod:
Did you initiate a dedicated repo for the stream, as you discussed one? If not then I can definitely start it and invite you guys.

jakoschiko:
I created the repo [tokio-maybe-tls](https://github.com/jakoschiko/tokio-maybe-tls/). I'll invite you once the CI is ready.

jakoschiko:
TIL https://docs.rs/tokio-util/latest/tokio_util/compat exists

PR #284, opened 2024-10-28
Update README.md

duesee:
This will vastly improve the fun/work ratio in this project. I'm open to other suggestions but it will be a tough discussion, @jakoschiko.


jakoschiko (approved):
I don't understand the concept of fun. But I'm willing to comply.

Issue #285, opened 2024-10-28
ci: Fix minimal-versions job


duesee:

jakoschiko:
@duesee Can we solve this similar to https://github.com/duesee/imap-codec/pull/615?

duesee:
I think so, yes.

PR #289, opened 2024-11-12
ci: Release with CI via git tag

### LINKED ISSUE #282: Release with CI via git tag

jakoschiko:
This workflow is similar to [the version in repository `imap-codec`](https://github.com/duesee/imap-codec/blob/main/.github/workflows/release.yml), but simpler because there is only a single crate that needs to be released.

Closes #282


jakoschiko:
We probably need to wait for #285. After that we can test the CI and then remove the `--dry-run`. And finally, @duesee, you would need to add a `secrets.CARGO_REGISTRY_TOKEN` to the repo.

duesee (approved):
Perfect (w/o the `--dry-run` :-))

PR #295, opened 2024-11-29
refactor: Use `AtomicUsize` instead of `AtomicU64`

### LINKED ISSUE #283: Use `AtomicUsize` instead of `AtomicU64`
See https://github.com/duesee/imap-codec/issues/603

duesee:
Closes #283 @jonassmedegaard, I've added you as a co-author. Are you fine with this? :-)


src/handle.rs
```diff
@@ -10,16 +10,16 @@ pub trait Handle {
...
 pub struct RawHandle {
-    generator_id: u64,
-    handle_id: u64,
+    generator_id: usize,
+    handle_id: usize,
 }
 
 impl RawHandle {
-    pub fn generator_id(&self) -> u64 {
+    pub fn generator_id(&self) -> usize {
```
duesee, line 18: This could be `pub crate`.

PR #296, opened 2024-11-29
Add option to discard greeting

### LINKED ISSUE #294: feat: Add `Client::new_without_greeting`
To allow a minimal (and secure) STARTTLS implementation, we need a way to start a session that skips the greeting. This allows us to handle the STARTTLS prefix in any way we please and continue with an isolated IMAP session after STARTTLS was already handled. This allows to fully isolate the plaintext and encrypted phase.

### Implicit TLS

```imap
<--- TLS handshake --->
// Startpoint `Client::tls(...)`
S: * OK ... // Greeting
C: ...
```

### STARTTLS

```imap
S: * OK ...
C: A STARTTLS
S: A OK ...
<--- TLS handshake --->
// Startpoint `Client::tls_no_greeting(...)`
// No greeting per spec.
C: ...
```

Do we want this to be a constructor or an `Option`? Any naming suggestions?

soywod:
Resolves #294.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -10,7 +10,6 @@ use imap_codec::{
     AuthenticateDataCodec, CommandCodec, GreetingCodec, IdleDoneCodec, ResponseCodec,
 };
 use thiserror::Error;
-use tracing::warn;
 
 use crate::{
     client_send::{ClientSendEvent, ClientSendState, ClientSendTermination},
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -73,6 +88,15 @@ impl Client {
         }
     }
 
+    pub fn discard_greeting(&mut self) {
```
duesee, line 91: Hm... I'm not sure I understand this. Do we need this? I'm asking because we already set the state via the `Options`? Also, as this merhod is `pub`, it may be called at unusual times and interfer with our state? Maybe I'm wrong.

soywod: I thought it could be useful in some cases. For example, when a Client is already initialized and STARTTLS is prompted later on (at runtime): it would be handy to just discard greeting rather than re-creating a new Client.

It does not cost that much to have setters, and it's a question of time someone will need it for a specific flow. Plus chances to break everything is low (as it needs to be done before receiving greetings). We could add a condition not to reset the state in case greeting is already discarded? Let me push a safer version.

That said, I would be fine not to add them as well, it was just a proposition. Let me know.

duesee:
I think I would prefer to go with a minimal version only via the `Options` for now. We can evaluate again when a usecase comes up.

soywod:
Fair enough, I just pushed the changes.

duesee (approved):
Thanks! :-)

